PR #20, opened 2021-01-14
feat(templates): add types for different template & wire choices through router/link

ramfox:
- Adds enum `TemplateType` to distinguish between each template
- Adds template objects that describe each template (this is what needs to be updated/altered when we come up with the template copy)
- Adds a `Templates` object that we can easily select each template from
- Adds `selectTemplate`, that pulls out a template based on the `TemplateType`
- In the `TemplateList` page, we now link to a dummy user page `/ds/test/dataset_${random num}`, that sends over the `TemplateType` for the workflow editor to parse
- The `WorkflowEditor` now has an effect that runs each time the location is updated: it checks if there is a specified template. If so, it pushes the specified template to the `WorkflowState`
- Also adds a `New Workflow` link at the top of navigation

### DIFF
--- a/frontend/src/features/template/TemplateList.tsx
+++ b/frontend/src/features/template/TemplateList.tsx
@@ -1,18 +1,23 @@
 import React from 'react'
 import { Link } from 'react-router-dom'
+import { TemplateType } from './templates'
 
 const templates = [
   {
-    name: 'CSV Download'
+    name: 'CSV Download',
+    type: TemplateType.CSVDownload
   },
   {
-    name: 'API Call'
+    name: 'API Call',
+    type: TemplateType.APICall
   },
   {
-    name: 'Database Query'
+    name: 'Database Query',
+    type: TemplateType.DatabaseQuery
   },
   {
-    name: 'Web Scrape'
+    name: 'Web Scrape',
+    type: TemplateType.Webscrape
   }
 ]
 
@@ -24,9 +29,12 @@ const TemplateList: React.FC<any> = () => {
       <br />
       <div className='text-lg text-bold flex flex-wrap -mx-2 overflow-hidden'>
         {
-          templates.map(({ name }) => (
-            <div className='my-2 px-2 overflow-hidden w-full md:w-1/2 lg:w-1/3 xl:w-1/3'>
-              <Link to='/datasets/edit'>
+          templates.map(({ name, type }) => (
+            <div key={name} className='my-2 px-2 overflow-hidden w-full md:w-1/2 lg:w-1/3 xl:w-1/3'>
+              <Link to={{ 
+                pathname: `/ds/temp/dataset_${Math.floor(Math.random() * 1000)}`,
+                state: { template: type }
+              }}>
                 <div className='border border-gray-300 hover:border-blue-500 rounded bg-white text-sm p-10 text-center'>
                     <h3>{name}</h3>
                 </div>
--- a/frontend/src/features/template/TemplateList.tsx
+++ b/frontend/src/features/template/TemplateList.tsx
@@ -1,23 +1,22 @@
 import React from 'react'
 import { Link } from 'react-router-dom'
-import { TemplateType } from './templates'
 
 const templates = [
   {
     name: 'CSV Download',
-    type: TemplateType.CSVDownload
+    type: 'CSVDownload'
   },
   {
     name: 'API Call',
-    type: TemplateType.APICall
+    type: 'APICall'
   },
   {
     name: 'Database Query',
-    type: TemplateType.DatabaseQuery
+    type: 'DatabaseQuery'
   },
   {
     name: 'Web Scrape',
-    type: TemplateType.Webscrape
+    type: 'Webscrape'
   }
 ]
 
@@ -32,7 +31,7 @@ const TemplateList: React.FC<any> = () => {
           templates.map(({ name, type }) => (
             <div key={name} className='my-2 px-2 overflow-hidden w-full md:w-1/2 lg:w-1/3 xl:w-1/3'>
               <Link to={{ 
-                pathname: `/ds/temp/dataset_${Math.floor(Math.random() * 1000)}`,
+                pathname: `/ds/me/dataset_${Math.floor(Math.random() * 1000)}`,
                 state: { template: type }
               }}>
                 <div className='border border-gray-300 hover:border-blue-500 rounded bg-white text-sm p-10 text-center'>
--- a/frontend/src/features/template/templates.ts
+++ b/frontend/src/features/template/templates.ts
@@ -1,0 +1,97 @@
+import { Workflow } from '../../qrimatic/workflow'
+
+export enum TemplateType {
+  CSVDownload = 'CSVDownload',
+  APICall = 'APICall',
+  DatabaseQuery = 'DatabaseQuery',
+  Webscrape = 'Webscrape'
+}
+
+export interface ITemplate extends Workflow {
+  type: TemplateType
+}
+
+// TODO (ramfox): need to formalize possible trigger & on completion types
+
+export const CSVDownload:  ITemplate = {
+  type: TemplateType.CSVDownload,
+  triggers: [{
+    // repeat every hour
+    type: 'cron',
+    value: 'R/PT1H'
+  }],
+  steps: [
+    { type: 'starlark', name: 'setup', value: `# load_ds("b5/csvdownload")` },
+    { type: 'starlark', name: 'download', value: `def download(ctx):\n\treturn "your download here"` },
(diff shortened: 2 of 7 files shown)


frontend/src/features/template/templates.ts
```diff
@@ -0,0 +1,97 @@
...
+export const Templates: {[key in TemplateType]: ITemplate} = {
+  [TemplateType.CSVDownload]: CSVDownload,
+  [TemplateType.APICall]: APICall,
+  [TemplateType.DatabaseQuery]: DatabaseQuery,
+  [TemplateType.Webscrape]: Webscrape
+}
+
+export function selectTemplate(templateType: TemplateType): ITemplate {
+  return Templates[templateType]
+}
```
ramfox: This felt like the simplest way to select a template. If anyone has better ideas please let me know.

frontend/src/features/workflow/state/workflowActions.ts
```diff
@@ -52,6 +53,18 @@ export function runEventLog(event: EventLogLine): EventLogAction {
   }
 }
 
+export interface SetWorkflowAction {
+  type: string
+  workflow: Workflow
+}
+
```
ramfox, line 60: @b5 wanted to get some clarification. There exists a `WorkflowAction`, but it is relatively specific
```
export interface WorkflowAction {
  type: string
  index: number
  value: string
}
```

Wasn't sure if I should have made that more general to fit multi-use cases (making it a general `WorkflowAction`) or if we should rename `WorkflowAction` to something more specific.

Went with creating a specific action for `SetWorkflowAction` as a middle ground.

frontend/src/features/workflow/state/workflowState.ts
```diff
@@ -73,4 +75,10 @@ function changeWorkflowStep(state: WorkflowState, action: WorkflowAction) {
 function addRunEvent(state: WorkflowState, action: EventLogAction) {
   state.events.push(action.data)
   state.events.sort((a,b) => a.ts - b.ts)
+}
+
+function setWorkflow(state: WorkflowState, action: SetWorkflowAction) {
+  state.workflow = action.workflow
+  state.events = []
+  return
```
ramfox, line 83: Can't describe how much more I like doing reducers this way

frontend/src/features/template/templates.ts
```diff
@@ -0,0 +1,97 @@
...
+  APICall = 'APICall',
+  DatabaseQuery = 'DatabaseQuery',
+  Webscrape = 'Webscrape'
+}
+
+export interface ITemplate extends Workflow {
+  type: TemplateType
+}
+
+// TODO (ramfox): need to formalize possible trigger & on completion types
```
ramfox, line 3: Personally feel like since we are only working with one trigger and one on completion type currently, we don't need to formalize this yet.

frontend/src/features/template/templates.ts
```diff
@@ -0,0 +1,97 @@
...
+  steps: [
+    { type: 'starlark', name: 'setup', value: `# load_ds("b5/csvdownload")` },
+    { type: 'starlark', name: 'download', value: `def download(ctx):\n\treturn "your download here"` },
+    { type: 'starlark', name: 'transform', value: 'def transform(ds,ctx):\n\tds.set_body([[1,2,3],[4,5,6]])' },
+    { type: 'save', name: 'save', value: '' }
+  ],
+  onCompletion: [
+    { type: 'push', value: 'https://registry.qri.cloud' }
+  ]
+}
```
ramfox: @chriswhong this is where the templates would get filled in. 

@b5 & @chriswhong  If a template doesn't require each step (setup, for example), should we still include it for now, until we get to a place where the user can add steps via UI?

frontend/src/features/template/TemplateList.tsx
```diff
@@ -24,9 +29,12 @@ const TemplateList: React.FC<any> = () => {
       <br />
       <div className='text-lg text-bold flex flex-wrap -mx-2 overflow-hidden'>
         {
-          templates.map(({ name }) => (
-            <div className='my-2 px-2 overflow-hidden w-full md:w-1/2 lg:w-1/3 xl:w-1/3'>
-              <Link to='/datasets/edit'>
+          templates.map(({ name, type }) => (
+            <div key={name} className='my-2 px-2 overflow-hidden w-full md:w-1/2 lg:w-1/3 xl:w-1/3'>
+              <Link to={{ 
+                pathname: `/ds/temp/dataset_${Math.floor(Math.random() * 1000)}`,
```
ramfox: Leaving the username `temp` for now. Happy to add an action that calls to the backend to get the session username, but that will only make sense in a local context, so wanted to check in before adding

frontend/src/features/template/templates.ts
```diff
@@ -0,0 +1,97 @@
+import { Workflow } from '../../qrimatic/workflow'
+
+export enum TemplateType {
+  CSVDownload = 'CSVDownload',
+  APICall = 'APICall',
+  DatabaseQuery = 'DatabaseQuery',
+  Webscrape = 'Webscrape'
+}
```
b5: We're going to have all these templates backed by a database at some point & dynamically fetched, which means we can't use a hard-coded enumeration. Let's use `id`s instead, and since workflows _should_ have an ID field, let's just add id to workflows and not use a `Template` type at all. The id strings themselves can be the same as the enumeration values.

frontend/src/features/template/templates.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export interface ITemplate extends Workflow {
+  type: TemplateType
+}
```
b5: If we were going to have a definition of a Template type, it should go in `src/qrimatic/template.ts`. Don't use the  `I` prefix on `ITemplate` (just `Template`).

frontend/src/features/template/templates.ts
b5: ok to skip for now, we should fix soon tho

frontend/src/features/template/templates.ts
b5: yes, include each step for now. Running scripts will behave weird if they're not all there b/c the backend is hard-wired to emit events that spit out all the steps. If we don't define all the steps in the workflow, the `workflow.steps` and `run.steps` arrays don't line up

frontend/src/features/template/templates.ts
```diff
@@ -0,0 +1,97 @@
...
+    { type: 'push', value: 'https://registry.qri.cloud' }
+  ]
+} 
+
+export const Templates: {[key in TemplateType]: ITemplate} = {
+  [TemplateType.CSVDownload]: CSVDownload,
+  [TemplateType.APICall]: APICall,
+  [TemplateType.DatabaseQuery]: DatabaseQuery,
+  [TemplateType.Webscrape]: Webscrape
+}
```
b5:
```suggestion
export const Templates: Record<string,Workflow> = {
  'CSVDownload': CSVDownload,
  'APICall': APICall,
  'DatabaseQuery': DatabaseQuery,
  'Webscrape': Webscrape
}
```

frontend/src/features/template/templates.ts
b5:
```suggestion
export function selectTemplate(id: string): Workflow {
  return Templates[id]
}
```

as a way to select templates, I'm all for it

frontend/src/features/workflow/state/workflowActions.ts
b5: agreed. `WorkflowAction` should be renamed to `SetWorkflowStepAction`

frontend/src/features/template/templates.ts
ramfox: Great.

So we already have a `datasetID` field, how do `id` and `datasetID` differ. I guess I'm basically asking what `datasetID` is referring to.

b5: `datasetID` refers to the `initID` of the dataset history we're writing to. This `id` is the _workflow_ identifier. It'll probably get stored on qrimatic

frontend/src/features/template/TemplateList.tsx
b5: yeah, this is great for now. even better if we use `me`, which will line up with backend conventions

frontend/src/features/template/templates.ts
ramfox: So we have at least 3 different ids, correct?
workflow ID -> id of the workflow
dataset ID -> id of the dataset
run ID -> id of the run, new after each run (found in each event)

Isn't there a one to one from workflow to dataset? What's the benefit of those being different?

b5: > Isn't there a one to one from workflow to dataset? What's the benefit of those being different?

It's not _always_ 1-1. The workflowID will be generated & stored by the scheduler (qrimatic). Not all datasets will be scheduled, so they'll have dataset IDs, but no workflowID. We'll associate triggers and completion hooks with workflowIDs

PR #23, opened 2021-01-15
feat(DatasetPreview): `SimpleTable` & `TabbedDisplay`

ramfox:
Screen Shot 2021-01-14 at 7 31 21 PM


frontend/src/qri/dataset.ts
```diff
@@ -1,5 +1,6 @@
 
 export interface Dataset {
+  [key: string]: any
```
b5: this degrades type safety. Instead identify the missing properties and add them to both the interface and the `newDataset` constructor function

frontend/src/qri/dataset.ts
```diff
@@ -213,3 +214,5 @@ export function newTransform(d: Record<string,any>): Transform {
 
 export type Body = Record<string, any> | any[][]
 
+export const ComponentList = ['commit', 'meta', 'structure', 'readme', 'body', 'transform', 'viz']
```
b5: using `List` as a suffix for an array type adds ambiguity. I'd prefer `ComponentNames`

frontend/src/qri/dataset.ts
```diff
@@ -142,7 +143,7 @@ export interface ColumnProperties {
 }
 
 export interface Schema {
-
+  [key: string]: any
```
b5, line 172: This, on the other hand is fine 😄

frontend/src/features/workflow/DatasetPreview.tsx
```diff
@@ -0,0 +1,155 @@
+import React, { useState } from 'react'
```
b5: one component per file please. I think it's also fine to add a directory within the feature folder:
```
/features
  /workflow
    /output
```
and throw all the output components in there

frontend/src/features/workflow/DatasetPreview.tsx
```diff
@@ -0,0 +1,155 @@
...
+      activeTab={componentTab}
+      tabs={tabs}
+      onTabClick={setComponentTab}
+      content={data[componentTab]}
+    />
+    <BodyDisplay 
+      structure={data.structure}
+      body={data.body}
+    />
+  </div>
```
b5: Body display should be one of the tabs, and selected by default.

frontend/src/features/workflow/DatasetPreview.tsx
```diff
@@ -0,0 +1,155 @@
...
+interface DatasetPreviewProps {
+  data: Dataset
+}
+
+export const DatasetPreview: React.FC<DatasetPreviewProps> = ({ data }) => {
+  const [componentTab, setComponentTab] = useState('commit')
+
+  const presentComponents: string[] = Object.keys(data).filter((c: string) => {
+    return ComponentList.includes(c) && c !== 'body'
+  })
```
b5:
```suggestion
  const presentComponents: string[] = Object.keys(data).filter((c: string) => ComponentList.includes(c) && c !== 'body')
```

frontend/src/features/workflow/DatasetPreview.tsx
```diff
@@ -0,0 +1,155 @@
+import React, { useState } from 'react'
+import { Dataset, ComponentList, Structure, Body } from '../../qri/dataset'
+
+interface DatasetPreviewProps {
+  data: Dataset
```
b5: I know we've used `data` as the "primary prop name" in the past, but I think that made much more sense when we had to write container mapping code and couldn't use selector hooks directly. We're already seeing places where multiple "data things" (stateful objects) are being passed as independent props with stuff like `<WorkflowCell step={step} run={run} />`.

For that, I think we should abandon the `data` prop convention and instead use lower-case names to match the data model whenever possible. With that convention in place, this prop should be `dataset: Dataset`

frontend/src/features/workflow/DatasetPreview.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+const TabbedDisplay: React.FC<TabbedDisplayProps> = (props) => {
+  const {
+    activeTab,
+    tabs,
+    onTabClick,
+    content
+  } = props
+
+  return <div className='w-full'>
```
b5: I'm fully bought in on the cool-kid way of writing this. Sorry for flip-flopping.
```suggestion
const TabbedDisplay: React.FC<TabbedDisplayProps> = ({
    activeTab,
    tabs,
    onTabClick,
    content
}) => (
  <div className='w-full'>
  {/* ... */}
```

frontend/src/features/workflow/DatasetPreview.tsx
```diff
@@ -0,0 +1,155 @@
...
+  ComponentList.forEach((component: string) => { 
+    if (component === 'body') {
+      return
+    }
+    const tab: Tab = { name: component }
+    if (presentComponents.includes(component)) {
+      tab.clickable = true
+    }
+    tabs.push(tab)
+  })
```
b5: instead of listing all components and greying out the undefined ones, let's only show components that are defined.

In addition, let's _not_ show the commit component, even if it exists. People seem to get confused when we tell them commit information is part of the dataset itself, and having one less component to think about saves space in user's brains.

Also confusing in this context: the transform component. Let's hide it as well for the time being.

This should also simplify the code. We could define a function in `qri/dataset.ts`:

```ts
export interface Component {
  name: string
  value: Record<string,any> | string // add the "string" for now to deal with readme component
}

export function newComponent(name: string, value: Record<string,any> | string): Component {
  return { name, value }
}

export function datasetComponents(ds: Dataset): Component[] {
  return componentNames.reduce((acc, name) => {
    if (ds[name]) {
      acc.push(newComponent(name, ds[name]))
    }
    return acc
  }, [])
}
```

note that the order of `componentNames` will still determine tab order. `commit` and `transform` dataset components should be removed with a `.filter((c: Component) => c.name !== 'transform' && c.name != 'commit')` in the `DatasetPreview` (react) component

frontend/src/features/workflow/DatasetPreview.tsx
```diff
@@ -0,0 +1,155 @@
...
+  </div>
+}
+
+interface BodyDisplayProps {
+  structure?: Structure
+  body?: Body
+}
+
+const BodyDisplay: React.FC<BodyDisplayProps> = (props) => {
+  const { structure, body } = props
```
b5: ditto on moving prop destructuring into the function signature

frontend/src/features/workflow/DatasetPreview.tsx
```diff
@@ -0,0 +1,155 @@
...
+    { structure.format !== 'csv'
+      ? <pre className='p-4'>{JSON.stringify(body, null, 2)}</pre>
+      : <SimpleTable {...props} />
+    }
+    <div className='text-right text-xs p-2'>Preview of first 10 rows</div>
+  </div>
+}
+
+const SimpleTable: React.FC<BodyDisplayProps> = (props) => {
+  const { structure, body} = props
```
b5: move prop destructuring into function signature

frontend/src/features/workflow/DatasetPreview.tsx
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  headers = [{"title":"", "type": ""}].concat(headers)
+  return <div className='overflow-auto max-h-100 border-t'>
+    <table className='table-auto border-collapse relative'>
+      <thead>
+        <tr>
+          {headers.map((header: Header, i: number) => {
+            return <th className='border-b border-l border-r text-left p-2 whitespace-nowrap sticky top-0 bg-white shadow' key={i}>{header.title}</th>
+          })}
```
b5: use parens instead of brackets & drop the return on one-liner arrow functions:
```suggestion
          {headers.map((header: Header, i: number) => (
            <th className='border-b border-l border-r text-left p-2 whitespace-nowrap sticky top-0 bg-white shadow' key={i}>{header.title}</th>
          ))}
```

frontend/src/features/workflow/DatasetPreview.tsx
```diff
@@ -0,0 +1,155 @@
...
+      </thead>
+      <tbody>
+        {body.map((row: any[], i: number) => {
+          return <tr key={i}>
+            <td className='border text-center p-2'>{i+1}</td>
+            {row.map((el: any, j: number) => {
+              return <td className='border text-left p-2 whitespace-nowrap' key={j}>{el}</td>
+            })}
+          </tr>
+        })}
```
b5: ditto, parens & no return

frontend/src/features/workflow/DatasetPreview.tsx
```diff
@@ -0,0 +1,155 @@
...
+    </table>
+  </div>
+}
+
+interface Header {
+  title: string
+  type: string
+}
+
+function selectHeaders(structure: Structure): Header[] | undefined {
```
b5: let's avoid using `select` when we're not actually selecting from the state tree. How about `schemaHeaders`?

b5 (approved):
hrm, tabbed display still shows "greyed out" instead of not showing components at all, but that's a _very_ easy fix to make in a follow up. Let's get this merged!

ramfox:
> hrm, tabbed display still shows "greyed out" instead of not showing components at all, but that's a _very_ easy fix to make in a follow up. Let's get this merged!

Oh no, I must have missed that comment somehow?

Will add in a follow up

PR #29, opened 2021-01-22
SPIKE: collection view & workflow editor features

b5:
This is a code spike, DO NOT MERGE


qrimatic/scheduler/cron.go
```diff
@@ -0,0 +1,287 @@
+// package scheduler schedules dataset and shell script updates
+package scheduler
```
b5, line 2: This  `scheduler` is the "core" of qrimatic. It defines Workflows, and runs triggers.

**Getting the `scheduler` package & `Job` struct renamed to `Workflow` should be the first thing we land.**

I'm not really that into this `scheduler` package name. Generally I think this package should define data structure interfaces for persistence, and serialization rules for workflows, triggers, hooks.

Right now it also defines the scheduler event loop and the function signature for delegating to a "qri repo worker". I think that's fine to stay here.

Ideally this repo _doesn't_ need to import `github.com/qri-io/qri/lib`, and can instead work with other communication packages like `event` & lower-level data structure packages like `dataset`. Not sure if we can fully pull that off tho

qrimatic/scheduler/cron.go
```diff
@@ -0,0 +1,287 @@
...
+	// WorkflowForName gets a workflow by it's name (which often matches the dataset name)
+	WorkflowForName(ctx context.Context, name string) (*Workflow, error)
+	// Workflow gets a single scheduled workflow by dataset identifier
+	Workflow(ctx context.Context, id string) (*Workflow, error)
+	// Runs gives a log of executed workflows
+	Runs(ctx context.Context, offset, limit int) ([]*Run, error)
+	// GetRun returns a single executed workflow by workflow.LogName
+	GetRun(ctx context.Context, id string, runNumber int) (*Run, error)
+	// // RunLogFile returns a reader for a file at the given name
+	// RunLogFile(ctx context.Context, id string, runNumber int) (io.ReadCloser, error)
```
b5, line 43: We've lost "log file" persistence, because those duties have moved to the event bus, and we don't have any event bus stuff wired up here. I think this can be handled next sprint (Sprint C) when we turn our attention to dataset activity feeds

qrimatic/scheduler/cron.go
```diff
@@ -0,0 +1,287 @@
...
+
+func init() {
+	golog.SetLogLevel("cron", "debug")
+}
+
+// Service is the generic interface for the Cron Scheduler, it's implemented
+// by both Cron and HTTPClient for easier RPC communication
+type Service interface {
+	// ListWorkflows lists currently scheduled workflows
+	ListWorkflows(ctx context.Context, offset, limit int) ([]*Workflow, error)
```
b5, line 33: We need a high level talk-to-core interface like this, much the same way `lib` uses `remote`, where the whole subsystem is wrapped in a primary interface with a good constructor function & default local implemnetation.

I'd love to get an endpoint up for this & use it to drive the collection view. We need to add the `VersionInfo` field to `Workflow` and cache it first.

We haven't yet figured out weather the qrimatic collection view should show datasets that _are not_ automated. I'd leave that as an issue for later, and instead just focus on shipping this method as the foundation for an API endpoint that can power the qrimatic collection view

qrimatic/scheduler/cron.go
```diff
@@ -0,0 +1,287 @@
...
+// workflow execution, and delegates the work of executing a workflow to a RunWorkflowFunc
+// implementation.
+type RunTransformFunc func(ctx context.Context, streams ioes.IOStreams, workflow *Workflow) error
+
+// RunWorkflowFactory is a function that returns a runner
+type RunWorkflowFactory func(ctx context.Context) (runner RunTransformFunc)
+
+// Cron coordinates the scheduling of running workflows at specified periodicities
+// (intervals) with a provided workflow runner function
+type Cron struct {
```
b5, line 61: This name is a carry over from the old `cron` package. It behaves a lot like `lib.Instance`, gathering up important subsystems that serve as arguments to business logic functions

qrimatic/scheduler/cron.go
```diff
@@ -0,0 +1,287 @@
...
+// 	}
+
+// 	if workflow.LogFilePath == "" {
+// 		return ioutil.NopCloser(&bytes.Buffer{}), nil
+// 	}
+
+// 	// TODO (b5): if logs are being stored somewhere other than local this'll break
+// 	// we should add an OpenLogFile method to LogFileCreator & rename the interface
+// 	return os.Open(workflow.LogFilePath)
+// }
```
b5, line 125: delete this

qrimatic/scheduler/cron.go
```diff
@@ -0,0 +1,287 @@
...
+
+type execTrigger struct {
+	w *Workflow
+	t Trigger
+}
```
b5, line 130: this struct is needed to bind together the trigger that actually fired the workflow with the workflow itself. We'll eventually want to send a structure like this through a channel, but I dislike this approach. Another alternative would be a private field on the `Workflow` struct for the trigger to fire.

Another subtle thing here: as written this relies on the trigger being a mutable pointer, _and_ that pointer being the same one from the `Workflow.Triggers` slice. Not super smart

qrimatic/scheduler/cron.go
```diff
@@ -0,0 +1,287 @@
...
+		select {
+		case <-t.C:
+			go check(ctx)
+		case <-ctx.Done():
+			return nil
+		}
+	}
+}
+
+func (c *Cron) runWorkflow(ctx context.Context, workflow *Workflow, t Trigger, runner RunTransformFunc) {
```
b5, line 182: The first thing this function should do is generate a `RunID`

qrimatic/scheduler/cron.go
```diff
@@ -0,0 +1,287 @@
...
+		}
+	}(workflow.Copy())
+
+	log.Debugf("run workflow: %s", workflow.ID)
+	if err := workflow.Advance(); err != nil {
+		log.Debugw("advancing workflow", "err", err, "id", workflow.ID)
+	}
+	if err := t.Advance(); err != nil {
+		log.Debugw("advancing trigger", "err", err, "id", t.Info().ID)
+	}
```
b5, line 195: I'm  a fan of this `Advance` method to isolate what actually happens to `Workflow` and `Trigger` state when a job runs (incrementing `RunCount`, setting `LastRunStart`, etc.) , but needing to  advance both independently is gross. Maybe we should be passing a `triggerIndex int` arg to this function instead and have `Workflow` call advance on it's own trigger.

qrimatic/scheduler/cron.go
```diff
@@ -0,0 +1,287 @@
...
+	if err := runner(ctx, streams, workflow); err != nil {
+		log.Errorf("run workflow: %s error: %s", workflow.ID, err.Error())
+		workflow.CurrentRun.Error = err.Error()
+	} else {
+		log.Debugf("run workflow: %s success", workflow.ID)
+		workflow.CurrentRun.Error = ""
+	}
+	now := NowFunc()
+	workflow.CurrentRun.Stop = &now
+	workflow.LatestRunStart = workflow.CurrentRun.Start
```
b5, line 216: All of this should be happening in `workflow.Advance()`

qrimatic/scheduler/trigger.go
```diff
@@ -0,0 +1,206 @@
...
+
+var (
+	TTCron TriggerType = "cron"
+)
+
+type Trigger interface {
+	Info() TriggerInfo
+	ToMap() map[string]interface{}
+	Advance() error
+}
```
b5, line 22: This trigger implementation should be a project on it's own.

I'm generally a fan of the way this whole Trigger implementation is shaping up. Some details to keep in mind:

* The distinct `Triggers` struct makes custom serialization a thing, even with distinct struct types for different "categories" of trigger
* The `TriggerInfo` is similar to os.FileInfo, but with a struct & fields instead of methods. It's also easier to embed in  Trigger implementations for common behaviour


In the past we've used `Type` in conjunction with a `Config map[string]interface{}` field. I think this solution is better in this case. Stronger typing. Moar compiler. better.

NEED TESTS. Tests for serialization, validation, and `Advance` in particular.

qrimatic/scheduler/cron.go
Arqu: I'm not sure on what the best format is for me to start carving out projects, so I'll latch onto places I think are good spots to drop a few lines. 

Anyways I think this is one of the cornerstone projects for the cloud version (and obviously same things go for local in this case) which is essentially just "write the API". Immediately I'd like to split that into 2 sub projects. One revolving any API design discussions and the other into the how it works (ie what else we need from lib and whether to augment core or to extend in qrimatic). Even stubs and interface definitions are useful here without the full logic coded out.

qrimatic/scheduler/cron.go
Arqu: We have this patter all over cloud. Usually the structure is called by the package name. Usually the packages have very different contexts of what they need and some shared elements. Ideally we can refactor to a more common struct once we get a bit closer to the mark.

qrimatic/scheduler/cron.go
Arqu: This feels like the Run Config we mentioned when talking about the spec (which we absorbed into the trigger config that's passed into the executor) I'm not against an explicit struct for this purpose instead of a hidden field because we might want to have override logic elsewhere for manual tweaking of the execution params ie tests and alike. Also maybe other stuff will use this too so rather export and reuse it than keeping it private.

qrimatic/scheduler/cron.go
Arqu: Personally I start getting the feeling we need to think about a proper state machine and flow setup to be at the heart of managing our job state and simply moving along the graph as state progresses. Lends itself nicely to properly modeling the flows of jobs but also simple latch points for transition events and emitting proper status updates in a central place, while having good recovery mechanics.

Which would also make a nice candidate for a project.

qrimatic/scheduler/http.go
```diff
@@ -1,4 +1,4 @@
-package cron
+package scheduler
```
Arqu, line 1: Think this should be the API package of qrimatic and we should have a layer deeper from this which would be what we now call the scheduler. Also the notion of a scheduler is blurred as it is definitely no longer only scheduling stuff. Proposing we move the package to something like `orchestrator` or `composer` or the like and having the scheduler being a piece of the puzzle dealing with the scheduler loop which essentially just ticks one of the many different types of triggers we will eventually have.

qrimatic/scheduler/store.go
```diff
@@ -0,0 +1,243 @@
...
+	"io"
+	"sync"
+)
+
+// ErrNotFound represents a lookup miss
+var ErrNotFound = fmt.Errorf("not found")
+
+// Store handles the persistence of Workflows and Runs. Store implementations
+// must be safe for concurrent use
+type Store interface {
```
Arqu, line 15: This is a great piece and the two discrete parts I see here are:
- Having the local file backed store
- Having the database powered cloud version of the store

qrimatic/scheduler/trigger.go
```diff
@@ -0,0 +1,206 @@
+package scheduler
```
Arqu, line 1: I would really like if we aimed to have webhook triggers (and on completes) very early on.
This would allow users for some fancy integrations on the cheap. Only thing I see that would take extra effort is working out the auth piece (but got that partially solved with cloud too - API key generation)

qrimatic/update/client.go
```diff
@@ -6,18 +6,17 @@ import (
 	"path/filepath"
```
Arqu, line 6: So this piece is my mental black box right now. I've been working on the deploy side of things and never really needed this part of qrimatic. I'm not sure I even understand what a qrimatic client would be. Is this just `lib` for CLI?

qrimatic/update/server.go
```diff
@@ -49,6 +49,7 @@ func NewServer(ctx context.Context, streams ioes.IOStreams, repoPath string) (*S
 func (s *Server) Serve(ctx context.Context) error {
 	apiServer := api.New(s.inst)
 	s.svc.AddRoutes(apiServer.Mux, apiServer.Middleware)
+	apiServer.Mux.Handle("/deploy", apiServer.Middleware(s.svc.NewDeployHandler("/deploy")))
```
Arqu, line 52: Again think we should separate out the API package as we did on cloud semi-recently and have this bellow be our `lib`

Arqu (commented):
There's a lot here. I did a first pass but intend to have another more thorough go at it when I can.

In terms of projects I can see the following areas (looking at backend primarily):
- API package - include all routes and handler definitions here, logically separated into named files
- Store package - definitions of the various data models and their cloud/local backed store implementations
- Lib package (or set of packages better)  - triggers (this includes the scheduling bit), on complete, workflow state machine, integration with core, other business & crud logic I can explicitly name yet 
- Events - think this should get isolated up and then be reused across the codebase
- The execution engine itself (local executor & external)

Hope this gives at least some input, I have to do another pass to elaborate some of the pieces more.

qrimatic/cmd/print.go
```diff
@@ -31,24 +31,24 @@ func printItems(w io.Writer, items []fmt.Stringer, offset int) error {
...
 	} else {
-		fmt.Fprintf(w, "%s\nin paused | %s\n", name(j.Name), j.Type)
+		fmt.Fprintf(w, "%s\nin paused\n", name(j.Name))
 	}
 	fmt.Fprintf(w, "\n")
 	return w.String()
 }
 
-type runStringer cron.Run
+type runStringer update.Run
```
b5, line 51: We need a way to print `Workflow` to the terminal for the `qrimatic list` command. This is ripped from qri core's `print.go` in the cmd

qrimatic/cron/cron.go
```diff
@@ -1,267 +0,0 @@
-// Package cron schedules dataset and shell script updates
-package cron
```
b5, line 2: this only looks deleted. Almost all of this is the result of refactoring the old `cron` package to now become the `scheduler` package. Lots of cruft is gone

qrimatic/scheduler/workflow.go
```diff
@@ -0,0 +1,218 @@
...
+)
+
+// workflowID returns CID string with a CronWorkflowCodecType prefix
+func workflowID(ownerID, dsID string) (string, error) {
+	mh, err := multihash.Encode([]byte(ownerID+"."+dsID), multihashCodec)
+	if err != nil {
+		return "", err
+	}
+	return cid.NewCidV1(WorkflowMulticodecType, multihash.Multihash(mh)).String(), nil
+}
```
b5, line 31: we agreed in a meeting to use UUIDs instead of this dumpster fire

qrimatic/update/client.go
b5: Yep you're totally right. This is _only_ for local debugging purposes. We shouldn't put much/any time into a client implementation quite yet

qrimatic/update/deploy.go
```diff
@@ -0,0 +1,83 @@
+package update
```
b5, line 1: The update package is kinda the `lib` of qrimatic, doing high level abstractions for the CLI & HTTP API.

Is the name terrible? yep. It's cruft from the old `update/cron` directory arrangement from when this stuff was in qri. It should be renamed.

We should also learn our lessons from qri where we wrote far too many tests in the `cmd` package, and instead write "integration-type" tests on THIS package. Exported functions here should map directly to API endpoints and CLI commands

b5:
this has served it's purpose. For the record we didn't enjoy working on massive code spikes like this 😓

PR #30, opened 2021-01-25
chore(storybook): configure storybook

b5:
add storybook to frontend


b5:
This isn't completely finished. Needs two things:

* wire up tailwind
* move the "run with" demo buttons into the Workflow Editor story

Even so I'm going to merge this b/c we at least get the `yarn storybook` command, and it actually does something.

cc @chriswhong & @ramfox

PR #35, opened 2021-01-26
refactor(frontend): NavBar is overridable, not required by default

b5:
Rework `NavBar` as a feature that other components can use. A no-props, no children `<NavBar />` element will generate a default app-level NavBar with a standard set of links. Providing children like this will override the "inside" of the NavBar (the main region of the NavBar between left & right corners:

```tsx
<NavBar>
  <h3>stuff in here</h3>
</NavBar>
```

There's an optional `menu` prop on NavBar that'll make a hamburger menu component show up in the  top left. Details on `NavBarMenuItem` type in the NavBarMenu component.

In the future when we add session support we'll turn `NavBar` into a container that selects the current user session if any and puts the profile photo / menu in the top right corner


chriswhong (approved):
LGTM

b5:
cc @ramfox for NavBar api & container planning

PR #36, opened 2021-01-26
feat: add dataset layout with side nav

chriswhong:
- Adds side nav with icons to the `Dataset` component
- Adds a two-column layout to the workflow editor to accommodate `WorkflowOutline`'
- Adds an `Icon` component (same implementation as Qri Desktop) for quickly adding FontAwesome Icons
- `App` is now always 100% of viewport height, scrolling should happen at lower levels when necessary

image


b5:
Looks good. I force-pushed some changes. Merging!

PR #37, opened 2021-01-26
refactor(frontend): send step-based transforms to backend for apply & deploy

b5:
update `Workflow` and `TransformStep` data types

### DIFF
--- a/frontend/src/features/template/templates.ts
+++ b/frontend/src/features/template/templates.ts
@@ -4,82 +4,91 @@ import { Workflow } from '../../qrimatic/workflow'
 
 export const CSVDownload: Workflow = {
   id: 'CSVDownload',
-  triggers: [{
+  datasetID: 'me/csv_download_template',
+  runCount: 0,
+  disabled: false,
+
+  triggers: [
     // repeat every hour
-    type: 'cron',
-    value: 'R/PT1H'
-  }],
+    { id: '', workflowID: '', type: 'cron', value: 'R/PT1H' }
+  ],
   steps: [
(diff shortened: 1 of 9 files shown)


frontend/src/features/template/templates.ts
```diff
@@ -4,82 +4,91 @@ import { Workflow } from '../../qrimatic/workflow'
...
-    { type: 'starlark', name: 'transform', value: 'def transform(ds,ctx):\n\tds.set_body([[1,2,3],[4,5,6]])' },
-    { type: 'save', name: 'save', value: '' }
+    { syntax: 'starlark', category: 'setup', name: 'setup', script: `# load_ds("b5/apicall")` },
+    { syntax: 'starlark', category: 'download', name: 'download', script: `def download(ctx):\n\treturn "your download here"` },
+    { syntax: 'starlark', category: 'transform', name: 'transform', script: 'def transform(ds,ctx):\n\tds.set_body([[1,2,3],[4,5,6]])' },
+    { syntax: 'qri', category: 'save', name: 'save', script: '' }
   ],
-  onCompletion: [
+  onComplete: [
     { type: 'push', value: 'https://registry.qri.cloud' }
```
Arqu, line 52: This is an interesting one. Does that mean that we push even on dry-run since it technically completed?

Arqu (approved):
I mean it looks good to me, no real objections.

frontend/src/features/template/templates.ts
b5: ah good point. It should really be `onSuccess`, shouldn't it?

PR #38, opened 2021-01-26
feat: add postCSS to storybook (for tailwind)

chriswhong:
Adds postcss webpack config to react storybook so that it can properly parse the `@tailwind base;` and other postCSS syntax in `index.css`

### DIFF
--- a/frontend/.storybook/main.js
+++ b/frontend/.storybook/main.js
@@ -1,3 +1,5 @@
+const path = require('path');
+
 module.exports = {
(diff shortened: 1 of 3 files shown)


ramfox (approved):
LGTM but we are getting a CI failure from something not from this commit?

```
src/features/workflow/state/workflowState.ts
  Line 5:10:  'NewWorkflow' is defined but never used  @typescript-eslint/no-unused-vars
```

PR #40, opened 2021-01-26
feat(`CodeEditor`): bring in MonacoEditor

ramfox:
first adjustments to layout

We have code highlighting!

Screen Shot 2021-01-26 at 6 31 50 PM

Two remaining (large) issues
1) first iteration of detection happens in the `editorDidMount` func however
it looks like the editor at this point doesn't actually have an accurate count
of how many lines are in the editor. Need to double check that it actually 
contains the correct content, otherwise the issue might be that before the 
component mounts we don't actually have the correct props/`script`
2) monaco editor seems to have some styling that forces a vertical scroll bar
or just forces a few extra empty "lines" at the end of the script so the scroll
bar is always present. Pretty sure we just need to figure out the correct 
settings to override


b5 (approved):
LGTM

PR #41, opened 2021-01-27
feat(frontend): WorkflowEditor loads transform referenced in url

b5:
Dataset component wraps the WorkflowEditor component. The dataset component extracts a qriRef from router params, and provides it to the WorkflowEditor. The Dataset component also fires off an API call to load the dataset that matches the ref whenever it changes.

Down in Workflow editor we fire an event to update the ref the workflow section of the state tree is tuned to with an effect hook. The workflow reducer listens for API_DATASET_SUCCESS actions and if the ref of the api response matches the ref the workflow reducer is tuned to, it updates the workflow transform script.

There are two bugs not handled yet:
* if the user has edited the script, it would get clobbered if a matching API fetch action completed while they were mid edit. The right solution there is to have *two* transform scripts in the workflow reducer. one we "reset" to, and another for user edits. successful api responses should write to the "reset" slot.
* the references don't confirm the version loaded is HEAD. successfully loading a historical commit will lead to the wrong script loaded into the workflow reducer

### DIFF
--- a/frontend/src/features/dataset/state/datasetActions.ts
+++ b/frontend/src/features/dataset/state/datasetActions.ts
@@ -1,9 +1,9 @@
-import Dataset, { newDataset } from "../../../qri/dataset";
+import Dataset, { NewDataset } from "../../../qri/dataset";
 import { QriRef } from "../../../qri/ref";
 import { ApiAction, ApiActionThunk, CALL_API } from "../../../store/api";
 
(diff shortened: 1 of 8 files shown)


b5:
@ramfox I'm going to merge this without review. Most important part is to read the description of what's going on, and the two bugs we'll need to cover. If you can turn both of those into an issue once you've read through I'd be grateful

PR #45, opened 2021-01-27
feat: layout improvements

chriswhong:
This PR has some layout improvements that will make it easier to navigate around the nascent app.  

- Adds `TopLevelLayout` component to standardize Navbar and content area for non-dataset views
- Improves `Splash` with some welcome text, subtitle, and buttons
- Standardizes height of Navbar in both contexts (dataset vs non-dataset)
- Adds a 'New Dataset' button to Navbar on non-dataset views
- Adds a two-line view to the name column in collection view, freeing up horizontal space and putting dataset metrics under the name"

React_Redux_App

### DIFF
--- a/frontend/src/chrome/RelativeTimestamp.tsx
+++ b/frontend/src/chrome/RelativeTimestamp.tsx
@@ -11,7 +11,7 @@ const RelativeTimestamp: React.FunctionComponent<RelativeTimestampProps> = ({ ti
     className='relative-timestamp'
     title={format(timestamp, 'MMM d yyyy, h:mm zz')}
   >
-    {formatDistanceToNow(timestamp)}
+    {formatDistanceToNow(timestamp, { addSuffix: true })}
   </span>
 )
 
--- a/frontend/src/features/collection/DatasetsTable.tsx
+++ b/frontend/src/features/collection/DatasetsTable.tsx
@@ -4,8 +4,8 @@ import ReactDataTable from 'react-data-table-component'
 import ReactTooltip from 'react-tooltip'
 import { Link } from 'react-router-dom'
 
+import Icon from '../../chrome/Icon'
 import RelativeTimestamp from '../../chrome/RelativeTimestamp'
-// import StatusIcons from './StatusIcons'
 import ExportButton from '../../chrome/ExportButton'
 import { VersionInfo } from '../../qri/versionInfo'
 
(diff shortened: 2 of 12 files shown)


frontend/src/chrome/TopLevelLayout.tsx
```diff
@@ -0,0 +1,18 @@
+// for non-dataset views, contains Navbar and full-height content area with
+//overflow-y-scroll
+
+import React from 'react'
+import NavBar from '../features/navbar/NavBar'
+
+interface TopLevelLayoutProps {}
```
b5: export this interface

frontend/src/chrome/TopLevelLayout.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+const TopLevelLayout: React.FunctionComponent<TopLevelLayoutProps> = ({ children }) => (
```
b5: Generally, I'm wary of this `Layout` pattern. The Layout components in desktop have caused a bunch of problems where different pages have had different requirements that lead to prop bloat.  Then layout problems start showing up & you end up playing whack-a-mole, fixing an issue in one screen only to have it create problems in another.

All that said, I think this time is different b/c there's no sidebar to contend with. But to  keep it that way I think we should rename this to `Page`, and do our best to avoid adding any props or special layout tricks here

frontend/src/chrome/RelativeTimestamp.tsx
```diff
@@ -11,7 +11,7 @@ const RelativeTimestamp: React.FunctionComponent<RelativeTimestampProps> = ({ ti
     className='relative-timestamp'
     title={format(timestamp, 'MMM d yyyy, h:mm zz')}
   >
-    {formatDistanceToNow(timestamp)}
+    {formatDistanceToNow(timestamp)} ago
```
b5: we don't always know this'll be in the past. Provide the `addSuffix` flag to the format function instead: 

```tsx
formatDistanceToNow(new Date(2014, 6, 2), { addSuffix: true })
```
docs: https://date-fns.org/docs/formatDistanceToNow

frontend/src/chrome/TopLevelLayout.tsx
b5: Use `FC` instead of `FunctionComponent`
```suggestion
const TopLevelLayout: React.FC<TopLevelLayoutProps> = ({ children }) => (
```

frontend/src/chrome/TopLevelLayout.tsx
```diff
@@ -0,0 +1,18 @@
+// for non-dataset views, contains Navbar and full-height content area with
+//overflow-y-scroll
```
b5: This comment should be above the `const` component declaration

frontend/src/chrome/TopLevelLayout.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import React from 'react'
+import NavBar from '../features/navbar/NavBar'
```
b5: Space btw package & local imports

frontend/src/features/collection/DatasetsTable.tsx
```diff
@@ -4,6 +4,7 @@ import ReactDataTable from 'react-data-table-component'
 import ReactTooltip from 'react-tooltip'
 import { Link } from 'react-router-dom'
 
+import Icon from '../../chrome/Icon'
 import RelativeTimestamp from '../../chrome/RelativeTimestamp'
 // import StatusIcons from './StatusIcons'
```
b5: delete unused import

frontend/src/features/notification/NotificationList.tsx
```diff
@@ -1,36 +1,42 @@
 import React from 'react'
 import { Link } from 'react-router-dom'
 import NotificationMenu from './NotificationMenu';
```
b5, line 4: Space btw package & local imports

frontend/src/features/splash/Splash.tsx
```diff
@@ -1,19 +1,37 @@
 import React from 'react'
 import { Link } from 'react-router-dom'
+import TopLevelLayout from '../../chrome/TopLevelLayout'
```
b5: Space btw package & local imports

b5 (approved):
Fantastic. A few nits to clean up then LGTM

b5:
made changes from review & force-pushed. merging so we have a nice place to start tomorrow!

PR #52, opened 2021-01-28
feat: monaco editor improvements

chriswhong:
- editor properly resizes width when the window is resized
- editor theme customizations (only sets gray background, but a custom theme is now defined in the code that we can to later)
- run bar appears above code editors (z-index change, was below)
- editor automatically increases vertical height when new lines are added
- editor no longer consumes scroll wheel events, so user can scroll up and down the workflow editor with ease


b5 (approved):
DON'T WAIT MERGE MEOW

PR #56, opened 2021-01-29
refactoring API

Arqu:
- Refactored the API stuff out into it's own package
- Cleaned up some code bits
- Added an `undeploy` route for a hard workflow remove/unschedule
- Added a way to update workflows which will also allow a rough way to manage triggers/hooks for now
- Fixed a deadlock on workflow deletes


qrimatic/update/server.go
```diff
@@ -16,11 +16,12 @@ import (
...
 	"github.com/qri-io/qri/lib"
 	"github.com/qri-io/qri/repo/gen"
+	qmapi "github.com/qri-io/qrimatic/api"
 )
 
 type Server struct {
 	inst *lib.Instance
-	svc  *Service
+	svc  *qmapi.Service
 }
```
b5: Here we have a filename `server.go` and `update.Server`, and a `qmapi.Service`. Let's switch the names of those two, keeping everything else the same

qrimatic/scheduler/store_file.go
```diff
@@ -287,3 +299,21 @@ func (s *FileStore) writeToFile() error {
 	}
 	return ioutil.WriteFile(s.path, data, 0644)
 }
+
+// Only use this when you have a surrounding lock
+func (s *FileStore) writeToFileNoLock() error {
```
b5, line 290: oh I like this pattern

b5 (approved):

PR #61, opened 2021-01-29
feat(`Workflow`): adjust UI based on the `lastRunType`

### LINKED ISSUE #57: bug(`WorkflowEditor`): when dry-running, "save" section should be "results"
If the run is a "dryrun" the "save" section should instead read "results"

Need to examine how we send over that we are doing "dry run" rather than a "run" and see if frontend can piggy back off of that. Need to make sure that once we "clear" it goes back to "save" (if we have a "clear" yet). At minimum, frontend knows that the user has clicked the "dry run" button and can set some state to say we want to call a specific section "results" instead of "save".

Hmm actually I think we get section names from the workflow steps. Is this a thing where were actually want backend to return a different "step" name? Or we can just interpret the output in context (which would be easier to just get done rn). Think we should interpret in context and bring up to backend team the option of returning a different step name.

ramfox:
closes #57 

- Adds `RunType` with options `dry` and `save`
- adds `getOutputNameFromRunType(runType: RunType): string` that takes a run type and converts it to a human readable name
- adds `lastRunType` and `setLastRunType` state to `Workflow` that tracks the LAST type of run we have done (defaulting to `RunType.dry`)
- plumbs `lastRunType` down into `WorkflowOutline`, and `WorkflowCell`, which uses `getOutputNameFromRunType` to determine what to show to the user.
- in `WorkflowCell`, adds description for "result"
- plumbs `setLastRunType` down to `RunBar`, when we click "run" or "dryrun", at that moment we should change the display (unless we "clear" the display, and then it should reflect if we have "run" or "dryrun" selected, but that's for the future)
- plumbs `setLastRunType` to the `deploy` modal as the param `setIfSave`. If we have "save" selected when we deploy, we need to ensure we need to update the `lastRunType`

Can test this by clicking the "deploy" button, which right now is defaulted to saving when clicked, and then clicking "run", which defaults to a dry run.

when we are able to switch between "dry run" and "run and save" WITHOUT HAVING RUN ANYTHING YET this will get more complicated, however until we can "clear" and also toggle between both states, we only need to rely on the last thing run.

### DIFF
--- a/frontend/src/features/workflow/WorkflowEditor.tsx
+++ b/frontend/src/features/workflow/WorkflowEditor.tsx
@@ -4,18 +4,22 @@ import { useDispatch } from 'react-redux';
 import WorkflowCell from './WorkflowCell';
 import WorkflowTriggersEditor from '../trigger/WorkflowTriggersEditor';
 import OnComplete from './OnComplete';
-import { NewRunStep, Run, RunState, RunStep } from '../../qrimatic/run';
+import { NewRunStep, Run, RunState, RunStep, RunType, getOutputNameFromRunType } from '../../qrimatic/run';
 import { TransformStep } from '../../qri/dataset';
 import { changeWorkflowTransformStep, runWorkflow } from './state/workflowActions';
 import RunBar from './RunBar';
 import { Workflow } from '../../qrimatic/workflow';
+import { SetRunTypeFunc } from './Workflow';
 
 export interface WorkflowEditorProps {
   workflow: Workflow
   run?: Run
+  lastRunType: RunType
+
+  setLastRunType: SetRunTypeFunc
 }
 
-const WorkflowEditor: React.FC<WorkflowEditorProps> = ({ run, workflow }) => {
+const WorkflowEditor: React.FC<WorkflowEditorProps> = ({ run, workflow, lastRunType, setLastRunType }) => {
   const dispatch = useDispatch()
 
   const [collapseStates, setCollapseStates] = useState({} as Record<string, "all" | "collapsed" | "only-editor" | "only-output">)
@@ -52,6 +56,9 @@ const WorkflowEditor: React.FC<WorkflowEditorProps> = ({ run, workflow }) => {
             <RunBar
               status={run ? run.status : RunState.waiting }
               onRun={() => {
+                // TODO (ramfox): when we have both "dry run" and "run and save"
+                // we need to set the `lastRunType` accordingly
+                setLastRunType(RunType.dry)
                 setCollapseStates({})
                 dispatch(runWorkflow(workflow))
               }}
@@ -69,7 +76,7 @@ const WorkflowEditor: React.FC<WorkflowEditorProps> = ({ run, workflow }) => {
             return (<WorkflowCell
               key={i}
               index={i}
-              step={step}
+              step={step.name === "save" ? {...step, name: getOutputNameFromRunType(lastRunType)} : step}
               run={r}
               collapseState={collapseState(step, r)}
               onChangeCollapse={(v) => {
--- a/frontend/src/features/workflow/WorkflowEditor.tsx
+++ b/frontend/src/features/workflow/WorkflowEditor.tsx
@@ -1,26 +1,25 @@
 import React, { useState } from 'react'
-import { useDispatch } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 
 import WorkflowCell from './WorkflowCell';
 import WorkflowTriggersEditor from '../trigger/WorkflowTriggersEditor';
 import OnComplete from './OnComplete';
-import { NewRunStep, Run, RunState, RunStep, RunType, getOutputNameFromRunType } from '../../qrimatic/run';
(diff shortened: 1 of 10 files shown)


frontend/src/features/workflow/WorkflowEditor.tsx
```diff
@@ -66,7 +71,7 @@ const WorkflowEditor: React.FC<WorkflowEditorProps> = ({ run, workflow }) => {
             return (<WorkflowCell
               key={i}
               index={i}
-              step={step}
+              step={step.name === "save" ? {...step, name: getOutputNameFromRunType(lastRunType)} : step}
```
ramfox: I'm a bit... inconsistent about when/where we should just replace the step name with "result". Here I do it at the `WorkflowEditor` level, deciding that below this line we understand what "result" is.

Another option is to rename the step "result" when we first pull it into the `Workflow`, and not have to worry about plumbing `lastRunType` down into lower components, but this would require iterating over the workflow steps and adjusting something we pulled from state... it felt weird, so I didn't do it.

frontend/src/features/workflow/modal/DeployWorkflowModal.tsx
```diff
@@ -1,21 +1,27 @@
...
 
   return (
     <div className="mt-3 sm:mt-0 sm:ml-4 sm:text-left">
      <h3 className="text-2xl leading-6 font-medium text-gray-900">Deploy</h3>
-     <p><input type='checkbox' checked disabled /> Run &amp; save before deploying</p>
+     <p><input type='checkbox' checked={save} onClick={() => setSave(!save)} disabled /> Run &amp; save before deploying</p>
      <button
        className='py-1 px-4 mx-1 font-semibold shadow-md text-white bg-gray-600 hover:bg-gray-300'
        onClick={() => {
+         save && setIfSave()
```
ramfox: like with `RunBar` and `onRun`, at the moment of execution we need to let the UI know that we are potentially changing up the expectation of what comes out of the run

b5:
I'm concerned we're introducing a little too much confusion with this PR.

Specifically: I'm confused by the `last` in `lastRunType`. To me it would make more sense to model the "upcoming" run type by moving the runType state tracking into the workbench state tree. By "upcoming" I mean what should happen when they push the "run" button. adding a `runType` field to the workflow state tree will help keep this consistent

I'm sure part of the reason for modeling `lastRunType` is to coordinate with the outline view any existing `RunState` object. To handle that, I think switching the run type should invalidate any existing `Run` field in the workflow state tree if `runType` changes

ramfox:
Okay great, so this is naming weirdness is part of the confusion I'm having, specifically because things are not fully set up the way we ultimately want them to be, and confusion about if/when we have a "cleared" state. 

It seems like the "runType" is sometimes going to have to model the "lastRun" and sometimes the "upcomingRun" and sometimes we need to coordinate with what the "dry run" or "save and run" button show, and sometimes we need to coordinate on "they clicked 'deploy' with the option to 'save'".

When we first arrive on the `Workbench` page, we have to display the "steps" in some way. They haven't been "run" yet, so the way we present the steps (result vs save) have to do with the default we show, aka "result" and showing the option to "dry run". In this context, we are modeling the "upcomingRun" & we need to coordinate with the "dry run/run and save" button. If the user clicks to change the context of the "run" button to "run and save", we should be displaying "save" in the cells.

We don't currently have a "run and save" button yet, so this isn't set up.

Then, when the user actually "runs" (either dry or not) or "deploys with save" (both of which we can do in the app currently), now we suddenly need to model "lastRunType", not the upcoming run, because we need to match whatever output we get from the run/deploy action.

So now we have the output content in each workflow cell shown in the UI.

If I have "dry run" output content in the cells and then the user switches from "dry run" to "save and run", then if the content isn't zero-ed out, we aren't showing the correct details. So dope, we should clear any existing Run. But we should also probably clear any existing run when the user edits any script fields, right?

The main wrinkle is when you click "deploy" w/ save. Now, even if the button selected has "dry run", we should be showing "save" rather than "result" in the output. Now the "runType" and the button that shows the run type are out of sync. If we clear when we are here, then we need to match the "runType" with whatever the "run" button shows.

seems to me:
- both switching from "dry run" to "run and save" and vice versa and clicking to actually "run" means we clear the output and change the "runType". 
- clicking "deploy with save" changes the "runType", but this shouldn't affect the "run" button (this is where the RunType and run button get out of sync)
- "clearing" the content in any way should match the "runType" to whatever type is selected in the "run" button

We just don't have all these elements in place yet, so I was having a hard time modeling/setting up for that eventuality.

I think we need least two params to correctly model this: 'runType' and 'runButtonType' (or something to that effect), one we rely on to show what we display in the outline/editor content, and one that we use to display what the 'run' button shows/does, & what we use to understand what to set 'runType' to when we have "cleared" the output content.

Both params should either live as local state in "Workflow" or in the workflow state tree. Changing the `runButtonType` should change the `runType` and clear the run. The only other time `runType` is changed is when we click "deploy w/ save". 

If I've overthought this and there is an easier path, please I would prefer that.

ramfox:
refactored based on previous comment!

ramfox:
This branch/pr is no longer relevant after #91

PR #84, opened 2021-02-01
fix: align fields across frontend & backend

b5:
a few important tweaks to get things to line up

### DIFF
--- a/frontend/src/qrimatic/workflow.ts
+++ b/frontend/src/qrimatic/workflow.ts
@@ -67,7 +67,8 @@ export interface WorkflowHook {
 export function NewWorkflowHook(data: Record<string,any>): WorkflowHook {
   return {
     type: data.type,
-    value: data.value
+    value: data.value,
+    remote: data.remote
   }
 }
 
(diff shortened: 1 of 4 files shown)


frontend/src/qrimatic/workflow.ts
```diff
@@ -67,7 +67,8 @@ export interface WorkflowHook {
 export function NewWorkflowHook(data: Record<string,any>): WorkflowHook {
   return {
     type: data.type,
-    value: data.value
+    value: data.value,
+    remote: data.remote
```
Arqu, line 71: `remote` should be part of push hook, but not all hooks, right? On the go side all of these are distinct based on the `type` denominator.

Arqu (approved):
Go go go :D

frontend/src/qrimatic/workflow.ts
b5: yep. if the remove value is undefined it'll be undefined on the hook. There are better ways to to this, but at least the code is in the right place for later refactoring

PR #86, opened 2021-02-01
feat(`Scroller`): add components to localize scrolling coordination

ramfox:
adds:
- `ScrollerState` that tracks the most recent scroller position & scrolled element
- `ScrollerAction` that adds `setScrollerPos` and `setScrollElement` actions
- `Scroller` - the container div that actually does the scrolling
- `ScrollTrigger` - when clicked, it starts a chain reaction that ends with the `Scroller` scrolling to show the associated `ScrollElement`
- `ScrollElement` - the element that we actually want to scroll to

First step in getting the outline and the document of the Workflow to interact smoothly!


b5 (approved):
The overall approach of this is great.

I Ended up rebasing & working a few changes in while was there. Aside from nits like always exporting `Props` interfaces, the biggest change is reworking `ScrollElement` to use native HTML semantics.

We call those old `<a>` tags anchors b/c they're intended to serve this exact purpose. We generally miss this "I'm an anchor in a page" feature because a tags are also hyperlinks. Switching ScrollElement to be an anchor tag with an `id` prop sets up the browser to auto-scroll for us to URLS like `https://rosebud.qri.cloud/ds/b5/dataset/workflow#download-step` once we have server-side rendering.

PR #87, opened 2021-02-02
layout for deploy spinner, WorkflowStatus, and WorkflowMessage

chriswhong:
This adds a stubbed-in interaction for the deploy step.  See gif in #team channel in discord.

- The full screen spinner is a child of `Dataset` so it can cover the side nav but *not* the top nav when visible
- `WorkflowStatus` will show different icon, message, colors, etc in the workflow outline depending on the deploy state of the workflow.  See mockups in Figma.
- `WorkflowMessage` is a toast that can appear at the bottom of the workflow outline.  When the deploy is successful, we can use animations to have this toast appear.

image


b5 (approved):
did a bunch of work to seat this PR properly, including adding a new `deploy` feature with it's own state tree & actions.

After Merging this I don't think the "deploying" fullscreen overlay will show up anymore, but the `deploy` button in the workflow editor should accurately reflect the state of deployment

PR #90, opened 2021-02-08
feat: refactor dataset components UI to use tailwind CSS

chriswhong:
Refactors DatasetComponent to use tailwind CSS.

- There is more finessing to be done, but this work yields a passable read-only set of components
- Structure/Schema in particular has a lot of editing UI and state that was not addressed in this PR.

image


b5 (approved):
ok, let's smash the merge button & cleanup with follow-ups

PR #91, opened 2021-02-08
feat(frontend): "dry run" and "save" script runner modes

b5:
should address some of the same questions as #61, but with a different approach that should compose a little better with the backend in the long run.

Screen Shot 2021-02-08 at 3 40 23 PM


b5:
@chriswhong this also replaces `qrilightblue` and `qriblue` with `qriblue` and `qrinavy`,  which match tailwind's 100-to-900 convention. I've added a diagram to a new ["Styles" page in Figma](https://www.figma.com/file/hDJCfq1ZovGwkac3WmmvRT/Qrimatic?node-id=569%3A676):

Screen Shot 2021-02-08 at 3 42 42 PM

chriswhong (approved):
LGTM!

frontend/src/features/workflow/state/workflowActions.ts
```diff
@@ -62,10 +63,22 @@ export function changeWorkflowTrigger(index: number, trigger: WorkflowTrigger):
   }
 }
 
-export function runWorkflow(w: Workflow): ApiActionThunk {
+export interface RunModeAction {
+  type: string
+  mode: 'apply' | 'save'
```
ramfox: `mode: RunMode` ??

frontend/src/features/workflow/state/workflowActions.ts
```diff
@@ -62,10 +63,22 @@ export function changeWorkflowTrigger(index: number, trigger: WorkflowTrigger):
...
+}
+
+export function setRunMode(mode: 'apply' | 'save'): RunModeAction {
```
ramfox: `mode: RunMode`?

frontend/src/features/workflow/RunBar.tsx
```diff
@@ -1,34 +1,66 @@
...
+          <div className='inline-block align-middle'>
+            <RunStateIcon state={status} size='md' />
+          </div>
+        </div>
+        <div className='flex-1 text-right'>
+          <DropdownButton
+            value={(runMode === 'apply') ? runModes[0] : runModes[1]}
+            onClick={() => {(status === RunState.running) ? handleCancel() : handleRun() }}
+            onChangeValue={(v: Option) => { 
+              if (v.id === 'apply') {
```
ramfox: Would add a `clearRun` function here to clear the run state, so it's obvious to the user that the content of the run isn't associated with the button shown.

ramfox (approved):
Woo. Removing "save and deploy" (for now) and putting the output of the dry run into the "transform" cell rather than a new "result" cell simplifies this ALOT.

I still think we are missing the "clear run" action when you switch from "Dry Run" to "Save and  Run". If you do a "Dry Run", you get the output of green checks (or red marks of doom) and then switch to "Save and Run", the only UI difference is there is now a "save" section in the outline (and the save cell, if you happen to be at that part of the document). If others think that is enough or this isn't big enough of an issue, I'm not dying on this hill, but it is something that is bothering me enough that I keep bringing it up.

I'm getting a `Line 8:10:  'TransformStep' is defined but never used  @typescript-eslint/no-unused-vars` error when I `yarn start`.

Also fully acknowledging my tendency to go OP on enum types hahah, just having a union of strings is enough.

There were two small spots that used `'apply' | 'save'` as a type rather than the already defined `RunMode`. Otherwise LGTM

b5:
thanks for the review @ramfox, I've added state-clearing-on-run-mode-change. Totally agree is should be addressed, but I think we can handle it in the workflow reducer logic instead of as a separate action.

PR #92, opened 2021-02-09
test(`e2e`): get functioning "Happy Path" test up and running

ramfox:
First, had to adjust the options for the `monaco editor`, it was inserting tabs rather than spaces, and those tabs were 4 spaces wide rather than the 2 spaces we were expecting. This was messing with our ability to manipulate monaco (typing into the monaco editor required creative backspacing and adding spaces back whenever you hit "enter"), so needed to fix it before moving forward with the e2e tests.

Some things to keep in mind when using puppeteer (and I made these notes in the e2e tests for future reference:
- most of the functions in puppeteer's api return promises
- if you are trying to get to the underlying dom element (that is wrapped in a promise), the best way is to pass it into `page.evaluate` and return what you need from there
- you cannot use optional chaining inside an `expect`, which is why need to always check if the element we tried to select in the page exists before returning. Best practice as far as I can see right now is if the element cannot be found on the page, throw an error so jest can catch it & typescript will expect that that variable beyond that point exists


frontend/src/features/workflow/CodeEditor.tsx
```diff
@@ -37,17 +37,21 @@ const CodeEditor: React.FC<CodeEditorProps> = ({ script, onChange }) => {
...
-    setTheEditor(editor)
+  useEffect(() => {
     if (theEditor) {
       const theModel = theEditor.getModel()
       if (theModel) {
         const currLineHeight = theModel.getLineCount()
         if (currLineHeight > 4) {
           handleSetLineCount(currLineHeight)
         }
+        theModel.updateOptions({ tabSize: 2, insertSpaces: true, indentSize: 2 })
```
b5, line 48: had to quickly research tabs vs spaces. Looks like spaces is preferred. TIL

b5 (approved):
Don't have the bandwidth for a proper full review, but I want e2e tests so I vote we rebase & merge

PR #93, opened 2021-02-09
feat: add activity list component

chriswhong:
Adds `ActivityList` component, used for a dataset-specific activity feed at `/ds/:username/:name/history`, and used in a global activity feed at `/activity`

Uses dummy data located in the associated react storybook story.  Dummy data schema needs some attention to make sure it has all we will need.

image


frontend/src/features/activity/stories/data/activity.json
```diff
@@ -0,0 +1,74 @@
...
+		"status": "success",
+		"statusMessage": "body and meta changed",
+		"runNumber": 10,
+		"runID": "b9d1211c-6a53-11eb-9439-0242ac130002",
+		"initID": "u2gr5gy6a6s5jq7fhpiicpi5htnqd65ybey7r3pmaou7uwtqqisq",
+		"username": "b5",
+		"name": "world_bank_population",
+		"path": "/ipfs/Qmebr6jDGqPViYe7pXtggNN3wJ6SiQehEpU8aZLg83AVpr",
+		"profileID": "QmSyDX5LYTiwQi861F5NAwdHrrnd1iRGsoEvCyzQMUyZ4W",
+		"bodySize": 282000,
```
Arqu: `bodySize`, `bodyRows`... feels like it maps to the stats "component" we also use in change reports etc.
It needs formalizing and cleanup, but I'd keep it as a sub object instead of one flat activity item.

```
{
	"type": "run",
	"timestamp": "2020-03-19T16:10:33.741276588-04:00",
	"status": "success",
	"statusMessage": "body and meta changed",
	"runNumber": 10,
	"runID": "b9d1211c-6a53-11eb-9439-0242ac130002",
	"initID": "u2gr5gy6a6s5jq7fhpiicpi5htnqd65ybey7r3pmaou7uwtqqisq",
	"username": "b5",
	"name": "world_bank_population",
	"path": "/ipfs/Qmebr6jDGqPViYe7pXtggNN3wJ6SiQehEpU8aZLg83AVpr",
	"profileID": "QmSyDX5LYTiwQi861F5NAwdHrrnd1iRGsoEvCyzQMUyZ4W",
	"stats": {
		"bodySize": 282000,
		"bodyRows": 2345678,
		"bodyFormat": "csv",
		"changeAmount": 0.45,
		"numErrors": 0
	}
}
```

Arqu (approved):
Looks pretty good to me, only a quick opinion on the `activity.json` layout.

frontend/src/features/activity/stories/data/activity.json
b5: This data structure is quickly evolving to be an aggregation/caching of details from a few different places: run data, a dsref, structure, and commit components. Given that it's an integration I'm a fan of using a pattern that started to emerge with `VersionInfo`, and flatten the aggregation into a struct with name prefixes that hint at the true source of the data.

b5: Given this is a thing for listing, keeping it as flat as possible makes everything _much_ easier on the frontend

b5 (approved):
I've tacked-on a bunch of technical pieces to this PR to get it ready for backend integration, which included a bunch of refactoring for the `LogItem` data model. Looking good!

PR #94, opened 2021-02-10
feat(`DatasetComponents`): bring Component View into the app!

### LINKED ISSUE #77: Add Component View
This is basically the dataset UI (including history) from Qri Desktop, where the user can explore each version.

ramfox:
closes #77 

Decided the cleanest/fastest way to port the `ds_components/DatasetComponents` component was to isolate it from the logic of fetching the dataset or setting the selected component, for now.

We may decide that we want it to encapsulate all the logic needed to work on its own, (especially when we get pagination involved and then need to pass down a function that allows us to fetch new pages of the body, and/or how to do editing although I have some thoughts about that). However, isolating it this way means we open an easier path to using the Component Viewer in different parts of the app or porting it to other projects in the future.

Some other changes:
- We needed to align the names of the `qriRef` refs and the `params` that we use in our router, otherwise our pattern of using `NewQriRef(useParams())` to get a `qriRef` from the route will not work
- adding a `datasetPath` file that contains a `pathToWorkflowEditor`, `pathToDatasetViewer`, `pathToActivityFeed`, used these functions to abstract the paths whenever we have to link to the dataset
-  any calls to `/ds/:username/:name` have been changed to `/ds/:username/:name/workflow` to keep the same functionality, but allow the path matching we need to show ui in other parts of the app. We are going to switch to having the Component Viewer as the default route that we navigated to when clicking on a dataset, but in order to isolate changes, this pr just doubles down on Workflow Editor as the default route
- adds `loadBody`, `fetchBody`, and adds the `API_BODY_SUCCESS` action to the reducer

### DIFF
--- a/frontend/src/features/dataset/Dataset.tsx
+++ b/frontend/src/features/dataset/Dataset.tsx
@@ -6,7 +6,7 @@ import { useParams } from 'react-router-dom';
 import { newQriRef } from '../../qri/ref';
 import Workflow from '../workflow/Workflow';
 import DatasetComponents from './DatasetComponents';
-import { loadDataset } from './state/datasetActions'
+import { loadBody, loadDataset } from './state/datasetActions'
 import NavBar from '../navbar/NavBar';
 import DatasetNavSidebar from './DatasetNavSidebar';
 import DatasetTitleMenu from './DatasetTitleMenu';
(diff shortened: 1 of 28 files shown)


frontend/src/features/dataset/Dataset.tsx
```diff
@@ -19,27 +19,28 @@ const Dataset: React.FC<any> = () => {
...
-          <Switch>
-            <Route path='/ds/:username/:dataset' exact><Workflow qriRef={qriRef} /></Route>
-            <Route path='/ds/:username/:dataset/components'><Redirect to={`${url}/body`} /></Route>
-            <Route path='/ds/:username/:dataset/components/:componentName'><DatasetComponents /></Route>
-            <Route path='/ds/:username/:dataset/history'><DatasetActivityFeed qriRef={qriRef} /></Route>
-          </Switch>
-        </div>
+        <Switch>
+          <Route path='/ds/:username/:name/workflow'><Workflow qriRef={qriRef} /></Route>
+          <Route path='/ds/:username/:name' exact><Redirect to={`${url}/workflow`} /></Route>
```
ramfox, line 39: I know this currently doubles down on having workflow be the default view, but it felt like too much in one pr to work through/change this behavior.

frontend/src/features/dataset/SideNavItem.tsx
```diff
@@ -13,7 +13,7 @@ export interface SideNavItemProps {
 
 const SideNavItem: React.FC<SideNavItemProps> = ({ id, icon, to, tooltip }) => {
   const { pathname } = useLocation();
-  const active = pathname === to
+  const active = pathname.includes(to)
```
ramfox, line 16: Because this was so specific, the `SideNavItem` for the Dataset Viewer was never selected (the `pathname` included the `component`, and the `SideNavItem` `to` param did not)

Generalized this and modified `/ds/:username/:name/workflow` as the route pointing to the Workflow editor. This has the added benefit of making it easier for us to switch over to the Dataset Viewer as the default route in the future.

frontend/src/features/dataset/state/datasetActions.ts
```diff
@@ -2,10 +2,16 @@ import Dataset, { NewDataset } from "../../../qri/dataset";
 import { QriRef } from "../../../qri/ref";
 import { ApiAction, ApiActionThunk, CALL_API } from "../../../store/api";
 
+export const bodyPageSizeDefault = 50
```
ramfox, line 5: arbitrary

rendering all these rows is what is causing major slow downs in our `BodyTable` component

frontend/src/qri/dataset.ts
```diff
@@ -26,6 +26,22 @@ export function qriRefFromDataset(dataset: Dataset): QriRef {
...
+    !ds.meta &&
+    !ds.structure &&
+    !ds.body &&
+    !ds.bodyPath &&
+    !ds.readme &&
+    !ds.transform &&
+    !ds.stats &&
+    !ds.viz)
+}
+
```
ramfox, line 44: getting closer and closer to a ts dataset package :)

frontend/src/store/api.ts
```diff
@@ -36,7 +36,7 @@ export interface ApiAction extends AnyAction {
     pageInfo?: ApiPagination
     // map is a function
     // map defaults to the identity function
-    map?: (data: object|[]) => any
+    map?: (data: any) => any
```
ramfox, line 39: This was too specific. Some map function need to specify fields in the given object (like we see in the `mapBody` function)

Plus it says "defaults to the identity function" but then wasn't the identity function :)

ramfox (commented):
self reviewing b/c of timing constraints.

Added logic to cover when a dataset is no longer loading, but still doesn't exist because it was fetched in error

PR #95, opened 2021-02-10
feat(*):  "surfability" improvements (add `DropdownMenu` & `RemoveDatasetModal`)

chriswhong:
- Adds composable `ModalLayout` with standardized icon, content, and buttons
- Uses `DropdownMenu` for all dropdowns (replaces `NavBarMenu`)
- Updates Collection view to include placeholder content for last run data and deployment status
- The 'Remove...' option is correctly wired to the `removeDataset` modal
- Dropdown Items take a `disabled` prop and will be grayed out, no highlight, etc 

TODO:  We need some kind of global state for dropdowns so the user can't open more than one at a time.


frontend/src/chrome/DurationFormat.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import moment from 'moment'
```
b5: nope. plz read: https://momentjs.com/docs/#/-project-status/

We're actively switching to `date-ins`, try [formatDuration](https://date-fns.org/v2.17.0/docs/formatDuration)

chriswhong: fixed, also refactored `moment` usage in `DatasetComponent`, removed `moment` from package.json, etc.

ramfox:
Hey, this is exciting and looks great @chriswhong !

These are my main comments in terms of stuff I noticed that may make it harder to refactor in the future:
- when adding something chrome-like (like a button or a modal, where it's going to be used by many different parts of the app), `onClick` or `onAction`-like functions should be self contained & not need extra input (unless its the click event itself, which it should pass through). Specially looking at `DropdownMenuItems` in this case, they took an `onClick` that took a general `itemProps` param. It's better to expect the parent component to compose all the needed information so that the lower component doesn't need to pass anything special to the given function.
-  This one is big, and I think will super help in the future: when mocking content, it's better to do all the mocking outside of the component (or all at the top where it can be co-located). This isolates all the logic we will have to eventually rip out and has the added benefit of making it super easy to slip the correct data structure in when the backend is ready to feed it to us. Again, it's all about isolating: everything inside the component in question shouldn't know its being fed dumb data & when it's exchanged with real data, the effort to make the real data work should be relatively low. 
- If for some reason there is no option but to mock it inside of the dataset itself, write a TODO about what's going on/what you are mocking and when we know we will be in a position to replace it. aka, "When backend is able to return log lists with run status information, we can remove this mocking" It took me a little bit to understand "Oh this data structure is being mocked!" and having a quick note would make understanding the code a lot quicker

Otherwise, in terms of general code-cleanliness, every other adjustment was mainly around type clean up (in one part you even left yourself a comment to the effect of "this variable should only be of type 'info' or 'danger', so you had the instinct that  there should only be two acceptable options, it just didn't show up in the interface). But! That's really minor clean up/alignment and the above points were the only major thing I saw in terms of structural habits.

Yay!

Also, re: your above TODO:

> TODO: We need some kind of global state for dropdowns so the user can't open more than one at a time.

Take a look at `Modal` & `DropdownMenu` to see one way to handle this without using state. Basically, we listen for a 'click' to happen in the document, & if it doesn't include the given div, we do some action. In this class, we close the modal or the. menu. The syntax is strange and VERY finicky (you've probably experienced the hell of event listeners at some point when working on desktop), but thought you might be interested in seeing how it's handled using hooks in modern react.

frontend/src/chrome/tests/DurationFormat.test.ts
```diff
@@ -0,0 +1,22 @@
...
+    expect(convertToDuration(60)).toBe('1m 00s')
+  })
+  test('check one hour', () => {
+    expect(convertToDuration(60 * 60)).toBe('1h 0m 00s')
+  })
+  test('check one day', () => {
+    expect(convertToDuration(60*60*24)).toBe('1d 0h 0m 00s')
+  })
+  test('check duration with multiple parts', () => {
+    expect(convertToDuration((60*60*24)+(60*60)+(60)+1)).toBe('1d 1h 1m 01s')
```
ramfox, line 20: we can debate exactly how we want durations displayed in the future, but now we have tests to make it easier to adjust!

frontend/src/features/collection/DatasetsTable.tsx
```diff
@@ -75,65 +110,148 @@ const DatasetsTable: React.FC<DatasetsTableProps> = ({
...
+        return (
+          <div className='flex flex-col'>
+            <div className='flex items-center mb-2'>
+              <div className='font-bold mr-2'>23</div>
+              <RunStatusBadge status={runStatus} small />
+            </div>
+            <div className='text-gray-500 text-xs'>
+              <Icon icon='clock' size='sm'/> <span><DurationFormat seconds={runDuration} /> | <RelativeTimestamp timestamp={new Date(timestamp)}/></span>
+            </div>
+          </div>
```
ramfox, line 186: one thing I specifically did not handle was the case where there is no (or a negative) duration.

TBH this pr feels like it does enough, and the follow up pr that actually integrates the backend responses into the collection view will be where we need to clean up that interaction.

frontend/src/features/dataset/modal/RemoveDatasetModal.tsx
```diff
@@ -0,0 +1,48 @@
...
+      .then((action: AnyAction) => {
+        if (action.type.includes('SUCCESS')) {
+          dispatch(loadDatasets())
+          return
+        }
+        // TODO (ramfox): when we add in toasts, this should be replaced with a
+        // error/failure toast
+        alert(`error removing dataset: ${action.payload.err.message}`)
+      })
+  }
```
ramfox: @b5 just wanted to draw your attention to this note
syntax is a little funky and wanted your opinion.

frontend/package.json
```diff
@@ -32,7 +32,6 @@
     "history": "^4.10.1",
     "install": "^0.13.0",
     "json-schema": "^0.3.0",
-    "moment": "^2.29.1",
```
b5, line 35: 👍

frontend/src/chrome/DropdownMenu.tsx
```diff
@@ -0,0 +1,97 @@
...
+export interface DropDownMenuItem {
+  // link is the url string that clicking this item can send the user
+  // link overrides `onClick`, ie if you have both link and onClick set, only
+  // the link will work
+  link?: string
+  onClick?: (e?: React.MouseEvent) => void
+  text: string
+  disabled?: boolean
+  // icon: options are found in the `Icon` component
+  icon?: string
```
b5, line 15: In a follow up we should covert the icon strings to a string union type & use it here instead of a string

frontend/src/chrome/tests/DurationFormat.test.ts
b5: OMG TESTS

frontend/src/features/activityFeed/CollectionActivityFeed.tsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react'
+import { LogItem } from '../../qri/log'
```
b5: space before local imports

frontend/src/features/app/modal/Modal.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { useCallback, useEffect, useRef } from 'react'
```
b5, line 1: delighted to see `modal` move to it's own feature folder

frontend/src/features/app/modal/ModalLayout.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { useDispatch } from 'react-redux'
+
+import Icon from '../../../chrome/Icon'
+import { clearModal } from '../state/appActions'
+
+
+export interface ModalLayoutProps {
+  title: string
+  type?: 'info' | 'danger' 
+  icon?: string
```
b5, line 13: another place that would benefit from an icon string union type

frontend/src/features/app/modal/ModalLayout.tsx
```diff
@@ -0,0 +1,87 @@
...
+  actionButtonText: string
+  action: () => void
+}
+
+const ModalLayout: React.FC<ModalLayoutProps> = (props) => {
+  const {
+    title,
+    type='info',
+    icon=undefined,
```
b5: the default value of an optional field is undefined. Remove

frontend/src/features/app/modal/ModalLayout.tsx
```diff
@@ -0,0 +1,87 @@
...
+  type?: 'info' | 'danger' 
+  icon?: string
+  actionButtonText: string
+  action: () => void
+}
+
+const ModalLayout: React.FC<ModalLayoutProps> = (props) => {
+  const {
+    title,
+    type='info',
```
b5, line 20: if you're defaulting to 'info' here, this property is in fact required. callers of the ModalLayout component should be required to pass the type.

frontend/src/features/app/modal/ModalLayout.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+export interface ModalLayoutProps {
+  title: string
+  type?: 'info' | 'danger' 
+  icon?: string
+  actionButtonText: string
+  action: () => void
+}
+
+const ModalLayout: React.FC<ModalLayoutProps> = (props) => {
```
b5: we've moved away from the `props` var + `const` destructing pattern. Destructure directly in the function declaration.

frontend/src/features/collection/Collection.tsx
```diff
@@ -5,31 +5,36 @@ import PageLayout from '../app/PageLayout'
 import { VersionInfo } from '../../qri/versionInfo'
 import DatasetsTable from './DatasetsTable'
 import { loadDatasets } from './state/collectionActions'
-import { selectCollection } from './state/collectionState'
+import { selectCollection, selectIsCollectionLoading } from './state/collectionState'
+import { SyncLoader } from 'react-spinners'
```
b5: move this import up

frontend/src/features/dataset/modal/RemoveDatasetModal.tsx
b5: Agreed. I'd love an issue for this. I've taken a stab at refactoring the `APIActionType` thunk to add a type argument. We should consider the promises returned by API middleware / actions as well.

frontend/src/qri/log.ts
```diff
@@ -17,5 +17,6 @@ export interface LogItem {
 
   bodySize: number
   bodyRows: number
+  bodyFormat: string
```
b5: be more specific: https://github.com/qri-io/dataset/blob/3b1209495e963e4dee9484fdf11ccd4effbb92a2/data_format.go#L41-L48

Add this to dataset.ts:
```ts

type BodyDataFormat =
  | "cbor"
  | "json"
  | "csv"
  | "xlsx"
  | ""         // unknown format
```

Then change structure.format & logItem.bodyFormat fields to be of type `BodyDataFormat`

b5 (approved):
A bunch of small nits, but looks really good! There are a bunch of things in this PR I'd like to make use of while working on the activity feed 😄

frontend/src/chrome/DropdownMenu.tsx
ramfox: Yesssss

PR #101, opened 2021-02-12
feat: navbar refactor, buttons, auth modals

chriswhong:
Refactors `NavBar` to be a layout component with `brand`, `left`, and `right` content areas.  This allows us to more easily compose NavBars depending on the current route and login state.

Refactors `PageLayout` to be a layout component with `top` and `main` content areas (so NavBar can be composed separately and passed in as `top` content.

Adds `Button` component to standardize buttons across the app.  Buttons can 
be implemented with `type` and `size` props.

Adds LogIn and SignUp modals as custom modal layouts that do not use the `ModalLayout`.  These include links to each other, links to TOS and Privacy, and a rudimentary `TextInput` component, but are not wired up to actions and do not yet have error handling UI.

Home___Qri_Cloud_and_ExternalLink_tsx_—___qri_qrimatic


frontend/src/chrome/Button.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+
+export type ButtonType = 'primary' | 'light' | 'danger' | 'dark'
```
b5: Lovely string union! I personally prefer them spread across lines when defined:

```ts
export type ButtonType =
  | 'primary'
  | 'light'
  // ...
```

frontend/src/chrome/Button.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export interface ButtonProps {
+  onClick?: () => void
+  type?: ButtonType
+  size?: 'md' | 'lg'
+  className?: string
+}
+
+const Button: React.FC<ButtonProps> = ({ type='primary', size='md', className, onClick, children }) => {
```
b5: split prop restructuring onto new lines as well

frontend/src/chrome/ExternalLink.tsx
```diff
@@ -2,21 +2,21 @@ import React from 'react'
...
+  to: string
   children: React.ReactNode
   className?: string
   tooltip?: string
 }
 
 export function openExternal (e: React.MouseEvent, href: string) {}
 
-export const ExternalLink: React.FunctionComponent<ExternalLinkProps> = ({ id, href, children, className, tooltip }) =>
+export const ExternalLink: React.FunctionComponent<ExternalLinkProps> = ({ id, to, children, className, tooltip }) =>
```
b5: spread prop destructing across multiple lines. I'm v into the `to` rename

frontend/src/features/app/PageLayout.tsx
```diff
@@ -1,16 +1,17 @@
...
+const PageLayout: React.FC<PageLayoutProps> = ({ top, main }) => (
   <div className='flex flex-col h-full bg-gray-100'>
-    <NavBar />
+    {top}
     <div className='flex-grow overflow-y-scroll bg-white'>
-      {children}
+      {main}
     </div>
   </div>
 )
```
b5, line 17: This is an over-abstraction. This components-as-props pattern turns the `PageLayout` component into a black box that requires the reader open it up to understand the DOM structure. 

If a component doesn't want a `NavBar` in the header it  shouldn't use the `PageLayout` component. Make a second layout component instead.

With Very limited exception we should alway pass children to components using React's built-in `<Component><Child></Component>` / `children` prop.

frontend/src/features/activityFeed/CollectionActivityFeed.tsx
```diff
@@ -3,20 +3,33 @@ import React from 'react'
...
-        </header>
-        <ActivityList log={activity as LogItem[]}/>
-      </div>
-    </PageLayout>
+    <PageLayout
+      top={top}
+      main={main}
+    />
   )
 }
```
b5: I think this code proves the point about the changes to the `PageLayout` API. We might want to consider renaming `PageLayout` to something more specific.

frontend/src/features/navbar/UserMenu.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
```
b5: move this to a new `sessionUser` feature folder (`/src/features/sessionUser`), rename to `sessionUserMenu`

frontend/src/features/app/modal/SignUpModal.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <img className="w-9 mx-auto mb-4" src="https://qri.cloud/assets/apple-touch-icon.png" alt="logo" />
+      <div className='text-2xl font-semibold mb-6'>Sign Up for Qrimatic</div>
+      <div className='w-64 mx-auto'>
+        <div className='mb-6'>
+          <TextInput
+            placeholder='Email'
+          />
+          <TextInput
+            placeholder='Password'
+          />
```
b5, line 28: This component is a great setup for thinking about validation next sprint 😄

frontend/src/features/app/modal/SignUpModal.tsx
```diff
@@ -0,0 +1,46 @@
...
+        </div>
+        <Button size='lg' className='w-full mb-6'>Continue</Button>
+
+        <div className='mb-3 text-gray-500' style={{ fontSize: '.7rem' }}>
+          By continuing, you agree to Qri's <ExternalLink to='https://qri.io/legal/tos'>Terms of Service</ExternalLink> & <ExternalLink to='https://qri.io/legal/privacy-policy'>Privacy Policy</ExternalLink>.
```
b5, line 33: Loving this `ExternalLink`

frontend/src/features/navbar/NavBar.tsx
```diff
@@ -1,30 +1,27 @@
...
-import Icon from '../../chrome/Icon'
-
+import { Link } from 'react-router-dom'
 
 export interface NavBarProps {
-  menu?: DropDownMenuItem[]
+  brand?: React.Component
+  left?: React.Component
+  right?: React.Component
 }
```
b5: This is another example of over-abstraction. The better way to handle this is with different types of `NavBar` components. Passing react components as Props. Seems to me what we're really after here are multiple variations of "`PageLayout`" (`DatasetPageLayout`/`NavPageLayout`/`FullscreenPageLayout`) that may make use of different Navbar components.

b5 (changes requested):

frontend/src/features/navbar/NavBar.tsx
chriswhong: Replaced NavBar with two variants one for Authenticated Users and one for Anonymous Users

frontend/src/features/app/PageLayout.tsx
chriswhong: Replaced PageLayout with two variants, one where there is a user and one for anonymous users.

frontend/src/features/activityFeed/CollectionActivityFeed.tsx
chriswhong: Now uses `UserPageLayout`

b5 (approved):
Okay, so I think this set of changes introduces a new bunch of concerns with regard to transitions in session states: when a user logs in / session is acquired, we'll need to reload the page under this paradigm.

Generally I think this `Layout` approach causes more problems than it solves, and I think the better solution is to move away from it, and toward "thickening" components like navbar into containers that can select from the state tree.

All that said, we can totally land this sort of refactor atop this existing PR, which adds modals we very much need. So I vote we merge this & iterate atop it

PR #102, opened 2021-02-17
feat(`/collection`): func `scheduler.ListWorkflowInfos`

ramfox:
The `/workflows/list` route returns the entire set of datasets and workflows, in a combined list of `WorkflowInfos`. `WorkflowInfos` are essentially `VersionInfos` with an added `ID` field that specifies the associated workflow id, if one exists.

TODO:
- [ ] add `ListWorkflowInfos` test. Not sure if this makes sense at this point, we would essentially be testing for behavior we know we want to change and isn't what we are working toward in the long run.
- [ ] add service when a run starts to keep track of runs
- [ ] `/workflows/running` - add cron.GetRunningWorkflows()
- [ ] add ability to return workflows in reverse chronological order based on commit time (so we can match up with datasets returned from qri core)
- [ ] this is a qri update but, we want the ability to return datasets in reverse chronological order based on activity time 
- [ ] how is qrimatic dealing with scope on the qrimatic side? Is adding a "username" param to the listing function enough scope for now? Locally this does not make sense because you can have datasets in your list that don't have your peername, but can we assume (for now) that qrimatic scopes based on "username" & you don' thave access to workflows that aren't in your namespace?

```go
// psuedo code for structure to keep track of when a run starts/ends
// struct should be created/exist on `cron`
type RunningWorkflows struct {
  bus event.Bus
  lock mutex.lock
  // stored in chronological order
  running []*Workflow
}

// Must be created before whatever runs the workflows starts, so we don't miss any events
// add running service to w/e struct 
New(ctx context): RunningWorkflows {
  // create bus from context
  // subscribe to ETWorkflowStarted & ETWorkflowCompleted events
    // ETWorkflowStarted calls addRunning
    // ETWorkflowCompleted calls removeRunning
  return rs
}

// addRun adds given run to END of list rs.runs & uses mutex lock O(c)
(rs) addRunning

// removeRun removes given run from list of rs.runs & uses mutex lock O(n)
(rs) removeRunning

// getRuns returns list of running workflows in reverse chronological order (aka latest running first) 0(n)
(rs) getRunningWorkflows()
// iterate through list, last first return reversed list (think it's overkill to create a Sort)
```

### things noticed/bugs:
- because frontend has no concept of session or identity yet, all workflows are returning
  as "me/datasetName". When we create the workflows, we use this as the datasetID
- currently frontend removes a dataset but not a workflow. Does "unschedule" mean delete? It looks like it does, but the language is confusing. This also does not delete the underlying dataset. Should we be relying on the qri endpoints for that, and have frontend send requests to "unschedule workflow" and "remove dataset", or should "unschedule workflow" auto remove any associated dataset?
- we have limit, offset in some places and offset, limit in others, what wins? 

change to `/collection`

### DIFF
--- a/qrimatic/api/cron.go
+++ b/qrimatic/api/cron.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"context"
 	"encoding/json"
 	"net/http"
 
@@ -8,10 +9,12 @@ import (
 	"github.com/qri-io/qrimatic/scheduler"
 )
 
+// StatusHandler is the qrimatic heath check
 func (s *Server) StatusHandler(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusOK)
 }
 
+// WorkflowsHandler returns a list of workflows
 func (s *Server) WorkflowsHandler(w http.ResponseWriter, r *http.Request) {
 	switch r.Method {
 	case http.MethodGet:
@@ -126,4 +129,17 @@ func (s *Server) RunHandler(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusInternalServerError)
 	w.Write([]byte("not finished"))
 	// c.runWorkflow(r.Context(), nil)
+}
+
+// WorkflowListHandler returns a list of WorkflowInfos, which include workflows
+// that do and do not include runs
+func (s *Server) WorkflowListHandler(w http.ResponseWriter, r *http.Request) {
+	data, err := s.sched.ListWorkflowInfos(context.Background(), s.inst, 25, 0)
+	if err != nil {
+		log.Errorf("error listing workflowInfos: %w", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Write([]byte(err.Error()))
+		return
+	}
+	apiutil.WriteResponse(w, data)
 }
--- a/qrimatic/api/cron.go
+++ b/qrimatic/api/cron.go
@@ -24,7 +24,7 @@ func (s *Server) WorkflowsHandler(w http.ResponseWriter, r *http.Request) {
 
 		js, err := s.sched.ListWorkflows(r.Context(), offset, limit)
 		if err != nil {
-			w.WriteHeader(http.StatusInternalServerError)
+			apiutil.WriteErrResponse(w, http.StatusInternalServerError, err)
 			return
 		}
 
@@ -58,8 +58,7 @@ func (s *Server) WorkflowsHandler(w http.ResponseWriter, r *http.Request) {
 	case http.MethodDelete:
 		name := r.FormValue("name")
 		if err := s.sched.Unschedule(r.Context(), name); err != nil {
-			w.WriteHeader(http.StatusInternalServerError)
-			w.Write([]byte(err.Error()))
+			apiutil.WriteErrResponse(w, http.StatusInternalServerError, err)
 			return
 		}
 	}
@@ -70,9 +69,9 @@ func (s *Server) WorkflowHandler(w http.ResponseWriter, r *http.Request) {
 	workflow, err := s.sched.WorkflowForDataset(r.Context(), dsID)
 	if err != nil {
 		if err == scheduler.ErrNotFound {
-			w.WriteHeader(http.StatusNotFound)
+			apiutil.WriteErrResponse(w, http.StatusNotFound, err)
 		} else {
-			w.WriteHeader(http.StatusInternalServerError)
+			apiutil.WriteErrResponse(w, http.StatusInternalServerError, err)
 		}
 		w.Write([]byte(err.Error()))
 		return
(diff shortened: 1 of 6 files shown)


qrimatic/api/routes.go
```diff
@@ -22,6 +22,7 @@ func (s *Server) AddRoutes(m *http.ServeMux, prefix string, mw func(http.Handler
 func (s *Server) AddCronRoutes(m *http.ServeMux, mw func(http.HandlerFunc) http.HandlerFunc) {
 	m.HandleFunc("/cron", mw(s.StatusHandler))
 	m.HandleFunc("/workflows", mw(s.WorkflowsHandler))
+	m.HandleFunc("/workflow/list", mw(s.WorkflowListHandler))
```
ramfox: So we already have a `/workflows` endpoint. This endpoint returns a list of workflows, not a union of datasets and workflows. We may want to keep this, so wanted to point this out before removing.

`qrimatic list` is similar: it returns a list of workflows to the cmd, should this be converted to a union of datasets and workflows

qrimatic/api/cron.go
```diff
@@ -127,3 +130,16 @@ func (s *Server) RunHandler(w http.ResponseWriter, r *http.Request) {
 	w.Write([]byte("not finished"))
 	// c.runWorkflow(r.Context(), nil)
 }
+
+// WorkflowListHandler returns a list of WorkflowInfos, which include workflows
+// that do and do not include runs
+func (s *Server) WorkflowListHandler(w http.ResponseWriter, r *http.Request) {
+	data, err := s.sched.ListWorkflowInfos(context.Background(), s.inst, 25, 0)
```
b5: use the request context instead of a background context, and flip the limit/offset to `offset`,`limit`:

```suggestion
	data, err := s.sched.ListWorkflowInfos(r.Context(), s.inst, 0, 25)
```

qrimatic/api/cron.go
```diff
@@ -127,3 +130,16 @@ func (s *Server) RunHandler(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		log.Errorf("error listing workflowInfos: %w", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Write([]byte(err.Error()))
```
b5: This should be `apiutil.WriteErrResponse(w, http.StatusBadRequest, err)`

qrimatic/scheduler/workflow_info.go
```diff
@@ -0,0 +1,108 @@
...
+			BodyFormat:  vi.BodyFormat,
+			NumErrors:   vi.NumErrors,
+			FSIPath:     vi.FSIPath,
+			RunID:       vi.RunID,
+			RunStatus:   vi.RunStatus,
+			RunDuration: vi.RunDuration,
+		},
+		ID: WorkflowID,
+	}
+}
```
b5: I don't think this function is necessary, or at least it doesn't need to be exported. Also, varNames shouldn't be capitalized (`WorkflowID` ->  `workflowID`)

I'd rewrite this as:
```go
func workflowInfoFromVersionInfo(vi dsref.VersionInfo, workflowID string) *WorkflowInfo {
  return &WorkflowInfo{
    VersionInfo: vi,
    ID: workflowID, 
  }
}
```

because VersionInfo isn't a pointer, it's pass-by-value, so this argument is already getting a copy of the VersionInfo struct. No need to hand-construct another

qrimatic/scheduler/workflow_info.go
```diff
@@ -0,0 +1,108 @@
...
+
+// ListWorkflowInfos returns all the WorkflowInfos (including datasets that have
+// been converted to WorkflowInfos). Future iterations will include pagination
+func (c *Cron) ListWorkflowInfos(ctx context.Context, inst *lib.Instance, after, before int) ([]*WorkflowInfo, error) {
+	m := lib.NewDatasetMethods(inst)
+	// TODO (ramfox): for now we are fetching everything.
+	p := &lib.ListParams{
+		Limit:  100,
```
b5: I have more than 100 datasets locally. Make this like `100000000000` (we _really_ need to support `-1` as a limit if we don't already)

qrimatic/scheduler/workflow_info.go
```diff
@@ -0,0 +1,108 @@
...
+	wiMap := map[string]*Workflow{}
+
+	for _, w := range ws {
+		wiMap[w.DatasetID] = w
+	}
+
+	wis := []*WorkflowInfo{}
+	for _, vi := range vis {
+		// DatasetID is currently `username/name`
+		viID := fmt.Sprintf("%s/%s", vi.Username, vi.Name)
```
b5:
```suggestion
		viID := vi.Alias()
```

b5 (approved):
a few nits, but good start!

qrimatic/scheduler/workflow_info.go
ramfox: ty, I thought I tried this first, but my syntax must have done something weird

qrimatic/scheduler/workflow_info.go
ramfox: fancyyyyy ty

qrimatic/api/cron.go
ramfox: Why is it a status bad request?

PR #107, opened 2021-02-19
feat: add mock session, modal improvements

chriswhong:
- wires up mock `session` state so users can experience the login/signup flow
- adds a dashboard at `/dashboard` which is where users are routed to after successful signup or login.
- adds `PrivateRoute`, which will redirect to /splash if the user is not logged in.
- brings in `formValidation.ts` from qri desktop, SignupModal performs input validation before firing its action
- adds ability to dismiss modals by pressing esc
- adds some icons to links to top-level views (collection, activity feed, dashboard)

Home___Qri_Cloud


b5 (approved):
Ended up adding a "review commit" that makes a bunch of tweaks to get things to gel with a little less overhead. @chriswhong def have a read through 64f4f9bbba6c91af2c0d03743042915788efd149 when you get the chance

PR #108, opened 2021-02-20
feat(`/collection/running`): get list of currently running WorkflowInfos to power the frontend `Collection` view

### LINKED ISSUE #103: feat(`scheduler.Store`): subscribe to events that should update workflows
We need to have the store listen to run and workflow events & update the store based on these events.

This will nicely delegate tasks between the different systems: cron and qri will emit events that indicate something is changing/happening, and the store understands how to update itself based on those events.

TODO:
* [x] construct a `TestRunWorkflow` test that ensure we emit the `ETWorkflowStarted`, `ETWorkflowCompleted` events & mutate the event payload's workflow properly
* [x] construct test that emits a stream of prepared workflow/run events & shows the store has properly "put" the workflows in the store
* [x] add `LatestStart`, `LatestStop`, and `Status` to Workflow
* [x] rename `Run` to `RunInfo`
* [x] Add `Subscribe` function that subscribes the workflow store to the workflow events & handles `workflow.PutWorkflow` 
* [x] In `RunWorkflow` function
    - before emitting ETWorkflowStarted (so in `workflow.Advance`)- should update `Workflow.LatestStart`, `Workflow.LatestEnd`, `Workflow.Status`
    - before emitting ETWorkflowCompleted - should update `Workflow.LatestEnd`, `WorkflowStatus`,
* [x] set up `api.RunningHandler` that `ListsWorkflowsByStatus(StatusRunning)` when the user fetches on `/collection/running`
* [x] set up `ListWorkflowsByStatus(status string)` that sends back a list of Workflows filtered by status, in reverse chronological order by `LatestStart`
* [x] set up `TestListWorkflowByStatus`

ramfox:
We need a way to get a snapshot of what qrimatic is currently running. We have the ability to listen to events over websockets so that frontend can catch as a workflow is changing, but there is no method for how to get a look into what was running before we made it into the app in the first place.

This closes #103 
- Adds `LatestStart`, `LatestEnd`, and `Status` to `Workflow`, to describe the `Workflow`'s own start, end, and status (that in the future will be separate from the run start, end, and status. Eventually these my be derived by a combination of `Run` and `OnComplete` events, but for now we are tracking the latest start, end, and status in the store)
- Adds `Subscribe` method to the Store that subscribes to emitted `Workflow` events, and updates the store accordingly
- Adjusts `cron.RunWorkflow` that emits the `ETWorkflowStarted` and `ETWorkflowCompleted` events, so that the accompanying workflows are adjusted correctly before they are emitted as event payloads
- Adds tests for the above systems
- Adds `ListWorkflowsByStatus` that returns a list of Workflows, filtered by `Status` & returned in reverse chronological order by `LatestStart`
- Adds endpoint `/collection/running` that returns a list of currently running `WorkflowInfo`s

probably easiest to look through `Files changed` at my self-review, or to look commit by commit


qrimatic/api/cron.go
```diff
@@ -138,3 +138,20 @@ func (s *Server) CollectionHandler(w http.ResponseWriter, r *http.Request) {
...
+	ws, err := s.sched.ListWorkflowsByStatus(context.Background(), scheduler.StatusRunning, 0, -1)
+	if err != nil {
+		log.Errorf("error listing currently running workflows: %w", err)
+		apiutil.WriteErrResponse(w, http.StatusBadRequest, err)
+		return
+	}
+	data := []*scheduler.WorkflowInfo{}
+	for _, w := range ws {
+		data = append(data, w.Info())
+	}
```
ramfox: tbh wasn't sure if I should push the conversion of `Workflows` to `WorkflowInfos` down lower? Maybe `cron` should have a `RunningCollection` function (rather than `ListWorkflowsByStatus`) that does this down in the `scheduler`?

qrimatic/api/routes.go
```diff
@@ -25,6 +25,7 @@ func (s *Server) AddCronRoutes(m *mux.Router, mw func(http.HandlerFunc) http.Han
 	m.HandleFunc("/cron", mw(s.StatusHandler))
 	m.HandleFunc("/workflows", mw(s.WorkflowsHandler))
 	m.HandleFunc("/workflows/trigger", mw(s.WorkflowManualTriggerHandler))
+	m.HandleFunc("/collection/running", mw(s.CollectionRunningHandler))
```
ramfox, line 28: nested it under `collection`, because it is returning `WorkflowInfos` rather than `Workflows` and it seemed logical to group the things that return `WorkflowInfos` meant to power the `collection` view together.

qrimatic/scheduler/cron.go
```diff
@@ -94,6 +94,12 @@ func (c *Cron) ListWorkflows(ctx context.Context, offset, limit int) ([]*Workflo
 	return c.store.ListWorkflows(ctx, offset, limit)
 }
 
+// ListWorkflowsByStatus proxies to the scheudler store for reading workflows by status
+// returns workflows is reverse chronological order by `LatestStart`
+func (c *Cron) ListWorkflowsByStatus(ctx context.Context, status string, offset, limit int) ([]*Workflow, error) {
+	return c.store.ListWorkflowsByStatus(ctx, status, offset, limit)
```
ramfox, line 100: Maybe this should be `ListWorkflowInfosByStatus` and the `Workflow` to `WorkflowInfo` conversion should happen here?

qrimatic/scheduler/cron_test.go
```diff
@@ -20,14 +22,11 @@ func mustRepeatingInterval(s string) iso8601.RepeatingInterval {
...
 func TestCronDataset(t *testing.T) {
 	updateCount := 0
-	next := time.Now().Add(time.Millisecond * 20)
 	workflow := &Workflow{
-		Name:         "b5/libp2p_node_count",
-		DatasetID:    "dsID",
-		OwnerID:      "ownerID",
-		Type:         JTDataset,
-		Periodicity:  mustRepeatingInterval("R/P1W"),
-		NextRunStart: &next,
```
ramfox, line 30: these fields don't exists any more

qrimatic/scheduler/deploy_test.go
```diff
@@ -1,90 +1,72 @@
...
-	cfg.Filesystems = []qfs.Config{
-		{Type: "mem"},
-		{Type: "local"},
-	}
-	cfg.Repo.Type = "mem"
+	// tr, err := repotest.NewTempRepo("foo", "deploy_test", repotest.NewTestCrypto())
+	// if err != nil {
+	// 	t.Fatal(err)
+	// }
+	// defer tr.Delete()
```
ramfox: This test was getting in my way when I was first iterating. I commented it out so I could move forward, but I could go back through (especially now that I have a way better understanding of this package) and get it to work again if that is a priority.

qrimatic/scheduler/http_test.go
```diff
@@ -1,82 +1,76 @@
...
+
+	// factory := func(context.Context) RunWorkflowFunc {
+	// 	return func(ctx context.Context, streams ioes.IOStreams, workflow *Workflow) error {
+	// 		return nil
+	// 	}
+	// }
+
+	// cliCtx := context.Background()
+	// cli := HTTPClient{Addr: ":7897"}
+	// if err := cli.Ping(); err != ErrUnreachable {
```
ramfox: Again, this was getting in my way when trying to iterate on tests. Plz let me know if I should delete wholesale or if commenting out is the best way forward rn.

qrimatic/scheduler/store.go
```diff
@@ -60,12 +73,63 @@ type MemStore struct {
...
+			return fmt.Errorf("error unexpected event: %q", e.Type)
+		}
+		return nil
+	}
+}
+
+// Subscribe allows the store to subscribe to workflow events that allow
+// the store to track and properly store updated `Workflows`
+func (s *MemStore) Subscribe(bus event.Bus) {
+	subscribe(s, bus)
```
ramfox: Not married to the way this is composed.

Seemed like a good idea to have the handler function exported if we ever want to create a `scheduler.Store` outside of this package & want it to have the same behavior.

Created the un-exported `subscribe` function as a convenience because we were doing the same thing in `MemStore` and `FileStore`

qrimatic/scheduler/store.go
```diff
@@ -87,12 +151,49 @@ func (s *MemStore) ListWorkflows(ctx context.Context, offset, limit int) ([]*Wor
...
+// chronological order by `LatestStart`
+func (s *MemStore) ListWorkflowsByStatus(ctx context.Context, status string, offset, limit int) ([]*Workflow, error) {
+	workflows := make([]*Workflow, 0, len(s.workflows.set))
+
+	for _, workflow := range s.workflows.set {
+		if workflow.Status == status {
+			log.Debugf("workflow %s has correct status", workflow.ID)
+			workflows = append(workflows, workflow)
+		}
+	}
```
ramfox, line 158: Hate that because of pagination & filtering we have to iterate over these lists multiple times.

RN O(n^2) run time worst case, yeesh

When we have a different kind of store backing the scheduler, we can get this to be less cumbersome.

Also, kept this as `offset, limit` rather than `after, before` to match `ListWorkflows`

That means this pr vs my last pr, I've treated pagination VERY differently. In the previous PR I used `after, before` as pagination place holders & purposefully ignored pagination in favor of returning the whole list. This pr, using `offset, limit` & not ignoring pagination at all. In the last pr we were working around some limitations of `qri`'s list filtering, and in this case we have full control over the filtering, which is why they were handled differently, but just wanted to point out the inconsistency.

qrimatic/scheduler/store_file_test.go
```diff
@@ -1,21 +1,99 @@
...
+	wID := w.ID
+	if err := store.PutWorkflow(ctx, w); err != nil {
+		t.Fatalf("putting workflow into store: %s", err)
+	}
+	expect := w.Copy()
+	expect.Status = "running"
+
+	bus.Publish(ctx, ETWorkflowStarted, expect)
+	// Better way to ensure we have given enough time for the store get the event & update?
+	<-time.After(time.Second)
```
ramfox, line 65: Any thoughts of a better way to ensure we have given the store enough time catch the event and update, or is this fine?

We can add another handler to the bus, but that would only let us know the event has emitted, not that the store itself has updated.

qrimatic/scheduler/store_file_test.go
```diff
@@ -1,21 +1,99 @@
...
+// a cmp.Diff of `iso8601.Duration`s
+func DurationCompare(x, y iso8601.Duration) bool {
+	return x.String() == y.String()
+}
+
+func TestSubscribe(t *testing.T) {
+	tmp, err := ioutil.TempDir(os.TempDir(), "TestStoreFile")
+	if err != nil {
+		t.Fatal(err)
+	}
```
ramfox, line 42: separating the concerns between the two subsystems (`cron` for manipulating the workflow as it changes emitting relevant events, and `store` for updating the store when necessary) made testing `cron` a bit more complicated (in terms of ensuring timing), but made testing this SO EASY.

It also really nicely isolated both sides of the problem. In general, I'm a fan!

qrimatic/scheduler/workflow.go
```diff
@@ -47,6 +47,15 @@ func workflowID() string {
...
 
+const (
+	// StatusRunning is the canonical constant for "running" execution state
+	StatusRunning = "running"
+	// StatusSucceeded is the canonical constant for "succeeded" execution state
+	StatusSucceeded = "succeeded"
+	// StatusFailed is the canonical constant for "failed" execution state
+	StatusFailed = "failed"
+)
+
```
ramfox: okay, my typescript brain was making this really hard for me to understand the best way to handle these.

feels like they should be their own type, but I know I tend to go overboard with types & enums in typescript world but... this should be a `WorkflowStatus` type right?

```
type WorkflowStatus string

const(
    StatusRunning WorkflowStatus = "running"
    StatusSucceeded = "succeeded"
    StatusFailed = "failed"
)
```

qrimatic/scheduler/workflow.go
```diff
@@ -107,15 +118,28 @@ func (workflow *Workflow) Complete(ds *dsref.Ref, ownerID string) error {
...
 	workflow.CurrentRun, err = NewRun(workflow.ID, workflow.RunCount+1)
 	if err != nil {
 		return err
 	}
 	workflow.RunCount++
-	workflow.LatestRunStart = workflow.CurrentRun.Start
+
+	workflow.LatestStart = workflow.CurrentRun.Start
+	workflow.LatestEnd = nil
+	workflow.Status = StatusRunning
```
ramfox, line 146: Felt like adjusting the workflow start, end, and status is exactly apart of what it means to `Advance` a workflow, but can pull this out into `RunWorkflow` if we want that to be a separate concern.

qrimatic/update/update_test.go
```diff
@@ -56,7 +56,7 @@ func TestDatasetWorkflowToCmd(t *testing.T) {
 		},
 	}
 	streams := ioes.NewDiscardIOStreams()
-	cmd := WorkflowToCmd(streams, dsj)
+	cmd := api.WorkflowToCmd(streams, dsj)
```
ramfox, line 59: this is just part of cleaning up tests so I could iterate on the tests I was writing for this pr.

qrimatic/scheduler/store.go
```diff
@@ -60,12 +73,63 @@ type MemStore struct {
...
+// and responds accordingly so that the store is always up to date
+func HandlerFromStore(s Store) event.Handler {
+	return func(ctx context.Context, e event.Event) error {
+		switch e.Type {
+		case ETWorkflowScheduled:
+			return nil
+		case ETWorkflowUnscheduled:
+			return nil
+		case ETWorkflowUpdated:
+			return nil
```
ramfox, line 105: felt out of the scope of this pr, but wondering if we should be mimicking the "let the store update itself" pattern on these other events.

qrimatic/scheduler/store.go
```diff
@@ -60,12 +73,63 @@ type MemStore struct {
...
+			}
+			log.Debugf("putting workflow %q with status %q into store", w.ID, w.Status)
+			s.PutWorkflow(ctx, w)
+		case ETWorkflowCompleted:
+			w, ok := e.Payload.(*Workflow)
+			if !ok {
+				log.Errorf("error expected event payload of %q to be of type %q", e.Type, "*Workflow")
+			}
+			log.Debugf("putting workflow %q with status %q into store", w.ID, w.Status)
+			s.PutWorkflow(ctx, w)
```
ramfox: hey can I get a room temperature of using `fallthrough` here? Both `ETWorkflowStarted` and `ETWorkflowCompleted` handle their events the same way. Only seen `fallthrough` in one or two places in our code bases before.

qrimatic/scheduler/workflow.go
b5: Yep exactly. `run.Status` is a good example to follow: https://github.com/qri-io/qri/blob/5484f838fafd05c604e5e777a728033f14199bb9/transform/run/run.go#L30-L51

You should define a distinct type here b/c workflows have fewer possible states

qrimatic/api/cron.go
b5: Yes exactly. Generally you want handlers as thin as you can possibly make them, as handlers are tightly coupled with the HTTP protocol. handlers should always call a matching business logic function like `RunningCollection`

qrimatic/scheduler/run.go
```diff
@@ -23,13 +23,13 @@ func runID(workflowID string, created time.Time) (string, error) {
 	return cid.NewCidV1(RunMulticodecType, multihash.Multihash(mh)).String(), nil
 }
 
-// Run is a record of workflow execution
+// RunInfo is a record of workflow execution
```
b5, line 26: At first I was a little suspicious of the need for a `RunInfo` struct, but I think it's right, we do need a way to represent runs without any logging data, and the need for a `WorkflowID` field means it can't exist within core

qrimatic/scheduler/run.go
```diff
@@ -23,13 +23,13 @@ func runID(workflowID string, created time.Time) (string, error) {
...
 // RunSet have one of three execution states, which describe it's position in
 // the execution lifecycle:
 // * unexected: workflow.RunStart == nil && workflow.RunStop == nil
 // * executing: !workflow.RunStart == nil && workflow.RunStop == nil
 // * completed: !workflow.RunStart == nil && !workflow.RunStop == nil
```
b5: Remove stale comment

qrimatic/scheduler/run.go
```diff
@@ -40,14 +40,14 @@ type Run struct {
 }
 
 // NewRun constructs a run pointer
-func NewRun(workflowID string, number int) (*Run, error) {
+func NewRun(workflowID string, number int) (*RunInfo, error) {
```
b5: `NewRun` -> `NewRunInfo`

qrimatic/scheduler/run.go
```diff
@@ -72,14 +72,14 @@ func (r *Run) Copy() *Run {
...
 // LogName returns a canonical name string for a run that's executed and saved
 // to a logging system
-func (r *Run) LogName() string {
+func (r *RunInfo) LogName() string {
 	return fmt.Sprintf("%s-%d", r.WorkflowID, r.Number)
 }
 
 // RunSet is a list of RunSet that implements the sort.Interface, sorting a list
 // of RunSet in reverse-chronological-then-alphabetical order
 type RunSet struct {
```
b5: RunSet -> RunInfoSet

qrimatic/scheduler/store.go
b5: yeah, save it for the future

qrimatic/scheduler/http_test.go
b5: This one is outdated & should be removed, especially since cron loses it's HTTP handler in this PR 😄

qrimatic/scheduler/deploy_test.go
b5: I'd like to see this test restored. It'll need tuning to bring it up to speed, but testing deploy is worth keeping

qrimatic/scheduler/http.go
```diff
@@ -167,7 +167,7 @@ func (c HTTPClient) Runs(ctx context.Context, offset, limit int) ([]*Run, error)
 }
 
 // GetRun returns a single executed workflow by workflow.LogName
-func (c HTTPClient) GetRun(ctx context.Context, logName string, number int) (*Run, error) {
+func (c HTTPClient) GetRun(ctx context.Context, logName string, number int) (*RunInfo, error) {
```
b5: rename this to `GetRunInfo`

qrimatic/scheduler/store.go
```diff
@@ -13,14 +16,24 @@ var ErrNotFound = fmt.Errorf("not found")
 // Store handles the persistence of Workflows and Runs. Store implementations
 // must be safe for concurrent use
 type Store interface {
+	// Subscribe allows the store to listen for workflow events so it can properly
+	// track and modify updated `Workflows`
+	Subscribe(bus event.Bus)
```
b5: I led you astray with this needing to be a method on the interface. I'd forgotten we have constructors that require a `event.Bus` argument. If you could, make two changes:

* remove the `Subscribe` method from the interface
* _un_-export `MemStore` and `FileStore`, forcing external callers to use the `NewMemStore` and `NewFileStore` constructors. Both should return a `Store` interface type

qrimatic/scheduler/store.go
b5: do it! It's rare, but when it works, it works

qrimatic/scheduler/store.go
b5: yeah, see my comment from above about preferring the unexported subscribe all the way. The right way to do this is to write a spec test that external implementations have to pass that would effectively require them to implement `HandlerFromStore`. We can leave `EventHandlerFromStore` exported for other implementations to use.

qrimatic/scheduler/store.go
```diff
@@ -60,12 +73,63 @@ type MemStore struct {
...
+		ETWorkflowUnscheduled,
+		ETWorkflowStarted,
+		ETWorkflowCompleted,
+		ETWorkflowUpdated,
+	)
+}
+
+// HandlerFromStore returns an `event.Handler` that listens to all Workflow events
+// and responds accordingly so that the store is always up to date
+func HandlerFromStore(s Store) event.Handler {
```
b5: rename to `EventHandlerFromStore`. Add a note that this is only for other store implementations and shouldn't need to be called on stores from this pacakge

qrimatic/scheduler/store.go
```diff
@@ -192,7 +293,7 @@ func (s *MemStore) DeleteWorkflow(ctx context.Context, id string) error {
 }
 
 // GetRun fetches a run by ID
-func (s *MemStore) GetRun(ctx context.Context, id string) (*Run, error) {
+func (s *MemStore) GetRun(ctx context.Context, id string) (*RunInfo, error) {
```
b5: GetRunInfo

qrimatic/scheduler/store.go
```diff
@@ -34,9 +47,9 @@ type Store interface {
 	// DeleteWorkflow removes a workflow from the store
 	DeleteWorkflow(ctx context.Context, id string) error
 
-	GetRun(ctx context.Context, id string) (*Run, error)
-	GetWorkflowRuns(ctx context.Context, workflowID string, offset, limit int) ([]*Run, error)
-	PutRun(ctx context.Context, r *Run) error
+	GetRun(ctx context.Context, id string) (*RunInfo, error)
+	GetWorkflowRuns(ctx context.Context, workflowID string, offset, limit int) ([]*RunInfo, error)
```
b5: `GetRun` -> `GetRunInfo`, `GetWorkflowRuns` -> `GetWorkflowRunInfos`

qrimatic/scheduler/store.go
```diff
@@ -34,9 +47,9 @@ type Store interface {
...
+	PutRun(ctx context.Context, r *RunInfo) error
```
b5: `PutRun` -> `PutRunInfo`

qrimatic/scheduler/workflow.go
b5: I agree this is the right place for this

qrimatic/scheduler/store_file.go
```diff
@@ -24,18 +27,26 @@ type FileStore struct {
...
+
 	log.Debugw("creating file store")
 	return s, s.loadFromFile()
 }
 
+// Subscribe allows the store to subscribe to relevant events that allow
+// the store to track and properly store updated `Workflows`
+func (s *FileStore) Subscribe(bus event.Bus) {
+	subscribe(s, bus)
+}
```
b5: same comment on removing this, calling the unexported subscribe in the constructor directly

qrimatic/scheduler/store_file_test.go
b5: I think this is fine for now, given that it's time-based. Coming up with a better setup should be a future endeavor

b5 (commented):
Looking very good! A few nits then LGTM

qrimatic/scheduler/run.go
ramfox: Okay full disclosure, you were the one who said rename `Run` to `RunInfo` 🤣

b5: lololololol my bad

ramfox: 😭 😭 😭 😭

b5 (approved):
haven't done a thorough reading, but this looks good to me. Let's merge & fix whatever crops up 😄

PR #113, opened 2021-02-26
feat(`Collection`): collection reflects currently running workflows

### LINKED ISSUE #110: feat(`Workflows`): adapt workflows reducer to include currently running workflows
* [x] add action for fetching `/collection`
* [x] add action for fetching `/collection/running`
* [x] adjust reducer:
    * [x] map w/ id: workflow of all workflows fetched
    * [x] list of wf ids that are running
* [x] when collection success, add wfs to workflow map, replacing any wfs that have been fetched before if the same id is encountered twice/previously
* [x] when collection_running success, add wfs to wf map, add ids to list of ids that are running
* [x] selectCollection: gets list of currently running ids, selects from the list, & orders from reverse chronological order last workflows first. iterate through rest of workflows, in reverse chronological order by commit time
* [x] catch WorkflowStarted and WorkflowCompleted events over the websocket, moving ids into and out of the "running" id list, and replacing the given workflowInfos with the most recent version

ramfox:
closes #110

You can test this by clicking "run now" and watching the workflow move to the top of the list & run time change.

There are still some questions about what we should be displaying to the user in terms of last commit time vs last workflow start or end.

There is also a larger questions about workflow status and how that interacts with the transform run status + the `onComplete` event status when we add on completes!


frontend/src/features/collection/WorkflowsTable.tsx
```diff
@@ -168,19 +158,25 @@ const WorkflowsTable: React.FC<WorkflowsTableProps> = ({
...
         return (
           <div className='flex flex-col'>
             <div className='flex items-center mb-2'>
               <div className='font-bold mr-2'>23</div>
-              <RunStatusBadge status={runStatus} size='sm' />
+              <RunStatusBadge status={status} size='sm' />
             </div>
             <div className='text-gray-500 text-xs'>
-              <Icon icon='clock' size='sm'/> <span><DurationFormat seconds={runDuration} /> | <RelativeTimestamp timestamp={new Date(timestamp)}/></span>
+              <Icon icon='clock' size='sm'/> <span><DurationFormat seconds={duration} /> | <RelativeTimestamp timestamp={new Date(commitTime || '')}/></span>
```
ramfox, line 179: we are currently showing the commit time, but it might make more sense to show the latest workflow end for datasets that have workflows

frontend/src/features/websocket/middleware/websocket.ts
```diff
@@ -80,6 +81,14 @@ const ETCronJobStarted = "cron:JobStarted"
 // payload is a Job
 // subscriptions do not block the publisher
 const ETCronJobCompleted = "cron:JobCompleted"
```
ramfox: I think `cron:JobCompleted` and `cron:JobStarted` may be dead code, but wanted to verify before removing all the reference to `Job` (rather than `Workflow`), because it crops up in a bunch of places in the frontend

qrimatic/scheduler/cron.go
```diff
@@ -253,9 +255,15 @@ func (c *Cron) RunWorkflow(ctx context.Context, workflow *Workflow, triggerID st
 	}
 
 	if err := runner(ctx, streams, workflow); err != nil {
-		log.Errorf("run workflow: %s error: %s", workflow.Name, err.Error())
-		workflow.CurrentRun.Error = err.Error()
-		workflow.Status = StatusFailed
+		if errors.Is(err, dsfs.ErrNoChanges) {
+			log.Debugf("run workflow: %s no changes", workflow.Name)
+			workflow.CurrentRun.Error = ""
+			workflow.Status = StatusNoChange
```
ramfox, line 261: Okay, so I'm assuming a few things here:
- `ErrNoChanges` should not be read as an error here
- we should be surfacing `NoChange` as a status on the `Workflow` (rather than only on the `Run`). My hesitation to surface it as an option for the `Workflow` is the conflict the `Run` status may have in the future with the `onComplete` status (what does it mean to have "no changes" but have a successful "on complete", and what do we show to the user?).

qrimatic/scheduler/workflow_info.go
```diff
@@ -83,7 +83,7 @@ func (c *Cron) ListCollection(ctx context.Context, inst *lib.Instance, before, a
 			wis = append(wis, w.Info())
 			continue
 		}
-		wis = append(wis, &WorkflowInfo{VersionInfo: vi})
+		wis = append(wis, &WorkflowInfo{VersionInfo: vi, ID: vi.Alias()})
```
ramfox, line 89: This is a bit of a hack until we get `InitID`s into `VersionInfo`s, as this value should probably be the `InitID`, rather than the alias.

frontend/src/features/collection/Collection.tsx
```diff
@@ -3,21 +3,34 @@ import { useDispatch, useSelector } from 'react-redux'
...
-import { loadCollection } from './state/collectionActions'
+import { loadCollection, loadRunningCollection } from './state/collectionActions'
 import { selectCollection, selectIsCollectionLoading } from './state/collectionState'
 import Page from '../app/Page'
 import WorkflowsTable from './WorkflowsTable'
 import { WorkflowInfo } from '../../qrimatic/workflow'
 import Button from '../../chrome/Button'
 import Icon from '../../chrome/Icon'
+import { getActionType } from '../../store/api'
+import { AnyAction } from '@reduxjs/toolkit'
```
b5: move import up to package imports group

frontend/src/features/collection/Collection.tsx
```diff
@@ -3,21 +3,34 @@ import { useDispatch, useSelector } from 'react-redux'
...
+        if (getActionType(action) === 'failure') {
+          console.log("load collection failed:", action.payload.err.message)
+        }
+        return await loadRunningCollection()(dispatch)
+      })
+      .then((action: AnyAction) => {
+        if (getActionType(action) === 'failure') {
+          console.log("loading running collection failed:", action.payload.err.message)
+        }
+      })
```
b5: this whole block taken together feels like the "real" `loadCollection` action.  Rename `loadCollection` into `loadCollectionWorkflows` and re-purpose `loadCollection`. Pseudocode:

```ts
export function loadCollection(dispatch: Dispatch, offset: number, limit:number) {
  loadCollectionWorkflows(1,50)(dispatch)
        .then(async (action: AnyAction) => {
          if (getActionType(action) === 'failure') {
            console.log("load collection failed:", action.payload.err.message)
          }
          return await loadRunningCollection()(dispatch)
        })
        .then((action: AnyAction) => {
          if (getActionType(action) === 'failure') {
            console.log("loading running collection failed:", action.payload.err.message)
          }
        })
}
```

Should remove a bunch of required imports

qrimatic/scheduler/workflow_info.go
b5: Agree, worth adding a `TODO(ramfox):`

frontend/src/features/collection/WorkflowsTable.tsx
b5: Agree. Let's keep this for now & switch once we have better tests up for workflow timestamps

frontend/src/features/collection/state/collectionActions.ts
```diff
@@ -22,7 +30,39 @@ function fetchCollection (page: number = 1, pageSize: number = 50): ApiAction {
...
+  }
+}
+
+export function workflowCompleted(workflow: Record<string, any>): WorkflowInfoAction {
+  const wf = NewWorkflow(workflow)
+  return {
+    type: WORKFLOW_COMPLETED,
+    data: workflowInfoFromWorkflow(wf)
+  }
 }
```
b5: can we get a comment above both of these pointing out that they're dispatched by the websocket, and users shouldn't need to dispatch them anywhere else? I keep looking at code like this & asking "wait, isn't this the server's job?" and re-remembering the websocket needs an action like this

frontend/src/features/websocket/middleware/websocket.ts
b5: yep, def ded code

qrimatic/scheduler/cron.go
b5: Let's start with the rule that a run that returns `ErrNoChanges` _cannot_ execute the `onComplete` step of a workflow. That keeps `no changes` in sync for both. In the future we might have to reconsider that

b5 (approved):
a few nits then LGTM

PR #114, opened 2021-02-26
feat: add web scraping template

chriswhong:
- updates the `Link` for template buttons to go to `/ds/me/dataset_xxx/workflow.  This needs to happen because the workflow editor is no longer the default dataset route, and the templates were not loading properly.
- modifies the "web scrape" template to include a simple example of scraping a unix timestamp from a website, parsing it, and creating a simple single-row dataset
image


chriswhong:
Closing as we will probably be changing these templates completely.

PR #115, opened 2021-02-26
refactor(`session`): remove "me" from the app, add the next layer of identity to frontend

ramfox:
* [x] session reducer
* [x] fetch from 'me' endpoint - `fetchUsername()` => takes the profile & returns the username of the node
* [x] disable signup & add error to inform the username
* [x] fetch the username inside the login action. If the username the user used to sign in is not the same as the one expected by the node, error
* [x] make sure there are TODOs saying that this is temporary
* [x] pull "username" from session & ensure we are using "username" everywhere instead of "me"
- TemplateList
- templates.ts
- workflowActions.ts
- NotificationsSettings.tsx
- datasetActions.ts

* [x] we need a new stand-in username for workflows that get created before the user has logged in. Using "new" for now. Eventually, when we keep track of the workflow inside the app, so that a user can refresh or navigate away without loosing changes, we should be changing the route from "new" to the given username after the user signs up. Right now, if we navigate to a new route, the page would refresh and any work that the user has done in the page will be erased.

### DIFF
--- a/frontend/src/features/dataset/Dataset.tsx
+++ b/frontend/src/features/dataset/Dataset.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect } from 'react'
-import { useDispatch } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 import { Redirect, Route, Switch, useRouteMatch } from 'react-router';
 import { useParams } from 'react-router-dom';
 
@@ -12,6 +12,7 @@ import DatasetNavSidebar from './DatasetNavSidebar';
 import DatasetTitleMenu from './DatasetTitleMenu';
 import DeployingScreen from '../deploy/DeployingScreen';
 import DatasetActivityFeed from '../activityFeed/DatasetActivityFeed';
+import { selectSessionUser } from '../session/state/sessionState';
 
 export interface DatasetMenuItem {
   text: string
@@ -21,6 +22,8 @@ export interface DatasetMenuItem {
 
 const Dataset: React.FC<any> = () => {
   const qriRef = newQriRef(useParams())
+  const username = useSelector(selectSessionUser)?.username || 'new'
+  qriRef.username = username
   const dispatch = useDispatch()
   const { url } = useRouteMatch()
 
--- a/frontend/src/features/dataset/Dataset.tsx
+++ b/frontend/src/features/dataset/Dataset.tsx
@@ -22,8 +22,16 @@ export interface DatasetMenuItem {
 
 const Dataset: React.FC<any> = () => {
   const qriRef = newQriRef(useParams())
-  const username = useSelector(selectSessionUser)?.username || 'new'
-  qriRef.username = username
+  const user = useSelector(selectSessionUser)
+
+  // This covers the case where a user created a new workflow before logging in.
+  // If they login while working on the workflow, the `user` will change, but the
+  // params used to generate the `qriRef` will not (because they are generated
+  // from the url, which has not changed). This check ensures that the correct 
+  // username is propagated after login/signup.
+  if (qriRef.username === 'new') {
+    qriRef.username = user.username
+  }
   const dispatch = useDispatch()
   const { url } = useRouteMatch()
 
--- a/frontend/src/features/dataset/state/datasetActions.ts
+++ b/frontend/src/features/dataset/state/datasetActions.ts
@@ -53,10 +53,7 @@ function fetchBody (ref: QriRef, page: number, pageSize: number): ApiAction {
         pageSize
       },
       segments: {
-        // TODO(b5): this is a hack for now while we don't have session support,
-        // restore ASAP
-        // peername: ref.username,
-        peername: 'me',
+        peername: ref.username,
         name: ref.name
       },
       map: mapBody
--- a/frontend/src/features/session/modal/LogInModal.tsx
+++ b/frontend/src/features/session/modal/LogInModal.tsx
@@ -2,9 +2,9 @@ import React, { useState } from 'react'
 import { Link } from 'react-router-dom'
 import { useDispatch, useSelector } from 'react-redux'
 import { SyncLoader } from 'react-spinners'
-import { getActionType } from '../../../store/api'
 import { AnyAction } from '@reduxjs/toolkit'
 
+import { ACTION_FAILURE, getActionType } from '../../../store/api'
 import ExternalLink from '../../../chrome/ExternalLink'
 import Button from '../../../chrome/Button'
 import TextInput from '../../../chrome/forms/TextInput'
@@ -30,7 +30,7 @@ const LogInModal: React.FC = () => {
   const handleButtonClick = () => {
     logIn(username, password)(dispatch)
       .then((action: AnyAction) => {
-        if (getActionType(action) === 'failure') {
+        if (getActionType(action) === ACTION_FAILURE) {
           setLoginError(action.payload.err.message)
           return
         }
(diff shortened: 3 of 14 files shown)


frontend/src/features/dataset/Dataset.tsx
```diff
@@ -21,6 +22,8 @@ export interface DatasetMenuItem {
 
 const Dataset: React.FC<any> = () => {
   const qriRef = newQriRef(useParams())
+  const username = useSelector(selectSessionUser)?.username || 'new'
+  qriRef.username = username
```
ramfox: we still need to be able to navigate to the `Dataset` aka workflow if there is no actual user. Once the user logs in, however, we need to make sure that we capture the new username. This won't change the route, but will make sure that the `Dataset` has the most up-to-date information. This will have to be adjusted when we have the ability to edit other people's workflows/datasets in qrimatic

ramfox:
Just a thought. We could refactor so we have these two routes that point to dataset:
`/ds/:name`
`/ds/:username/:name`

The former would assume that the username is the session username, the later would have no assumptions.

frontend/src/features/dataset/Dataset.tsx
b5: Interesting. Can we get a comment:

```
// TODO(ramrod): This will have to be adjusted when we have the ability to edit other people's workflows/datasets 
```

I think this is the right way to work around it for now, with one exception: the `|| 'new'` should be handled within the session reducer, not up in the component. Any reason we can't have the default state of the session user be a user with the username `new`?

frontend/src/features/dataset/state/datasetActions.ts
```diff
@@ -53,10 +53,7 @@ function fetchBody (ref: QriRef, page: number, pageSize: number): ApiAction {
         pageSize
       },
       segments: {
-        // TODO(b5): this is a hack for now while we don't have session support,
-        // restore ASAP
-        // peername: ref.username,
-        peername: 'me',
```
b5, line 59: yey ded TODO

frontend/src/features/session/modal/LogInModal.tsx
```diff
@@ -2,6 +2,8 @@ import React, { useState } from 'react'
 import { Link } from 'react-router-dom'
 import { useDispatch, useSelector } from 'react-redux'
 import { SyncLoader } from 'react-spinners'
+import { getActionType } from '../../../store/api'
```
b5: move into local imports group

frontend/src/features/session/modal/LogInModal.tsx
```diff
@@ -25,9 +28,14 @@ const LogInModal: React.FC = () => {
   }
 
   const handleButtonClick = () => {
-    dispatch(logIn(username, password)).then(() => {
-      dispatch(clearModal())
-    })
+    logIn(username, password)(dispatch)
+      .then((action: AnyAction) => {
+        if (getActionType(action) === 'failure') {
```
b5: it'd be great if `'failure'` was a constant defined in the api middleware. Don't need to do it this PR, there are a few places that need to switch on API response failure, calling them out with constants will help down the road

frontend/src/features/session/modal/SignUpModal.tsx
```diff
@@ -1,6 +1,8 @@
 import React, { useState } from 'react'
 import { useDispatch, useSelector } from 'react-redux';
 import { SyncLoader } from 'react-spinners'
+import { getActionType } from '../../../store/api';
```
b5: move into local imports group

frontend/src/features/session/modal/SignUpModal.tsx
```diff
@@ -9,18 +11,20 @@ import { showModal, clearModal } from '../../app/state/appActions'
...
 import { selectIsSessionLoading } from '../state/sessionState'
-import { validateEmail, validateUsername, validatePassword } from '../state/formValidation'
+import { validateEmail, validateUsername, validatePassword, ValidationError } from '../state/formValidation'
 
 const SignUpModal: React.FC = () => {
 
-  const [ email, setEmail ] = useState('')
-  const [ emailError, setEmailError ] = useState(null)
+  const [email, setEmail] = useState('')
+  const [emailError, setEmailError] = useState<ValidationError>(null)
```
b5, line 19: great addition

frontend/src/features/session/modal/SignUpModal.tsx
```diff
@@ -43,9 +47,14 @@ const SignUpModal: React.FC = () => {
     setPasswordError(passwordError)
 
     if (!emailError && !usernameError && !passwordError) {
-      dispatch(signUp(email, username, password)).then((action) => {
-        dispatch(clearModal())
-      })
+      signUp(email, username, password)(dispatch)
+        .then((action: AnyAction) => {
+          if (getActionType(action) === 'failure') {
```
b5: another place where an api failure constant would help

frontend/src/features/session/state/sessionActions.ts
```diff
@@ -1,17 +1,40 @@
-import { ApiActionThunk} from '../../../store/api'
+import { AnyAction } from '@reduxjs/toolkit'
+import { ApiAction, ApiActionThunk, CALL_API} from '../../../store/api'
```
b5, line 3: add a line to separate imports into package & local groups

frontend/src/routes.tsx
```diff
@@ -41,6 +41,7 @@ export default function Routes () {
         <PrivateRoute path='/collection'><Collection /></PrivateRoute>
         <PrivateRoute path='/activity'><CollectionActivityFeed /></PrivateRoute>
 
+        <Route path='/ds/new/:name'><Dataset /></Route>
```
b5, line 44: big fan of the separate route to support unauthenticated access

frontend/src/features/template/TemplateList.tsx
```diff
@@ -23,6 +25,7 @@ const templates = [
 ]
 
 const TemplateList: React.FC<any> = () => {
+  const username = useSelector(selectSessionUser)?.username || 'new'
```
b5: another spot where having the reducer always handle the `new` case would catch a potential bug if we don't update this in both places

frontend/src/features/session/state/sessionActions.ts
```diff
@@ -27,18 +50,34 @@ export function logOut (): ApiActionThunk {
 }
 
 export function signUp (email: string, username: string, password: string): ApiActionThunk {
+  // TODO (ramfox): until we have a proper sign up process (that takes a password)
+  // and allows the user to have its own identity on the qri node that is separate
+  // from the id assocaited with the qri node itself, let's disable this 
+  // the failure dispatches a message to the user saying that there is no
+  // way to currently signup
```
b5, line 58: 👍

b5 (approved):
Some small nits. If the session reducer can handle `new`, I say let's get this cleaned up & merged

b5:
> Just a thought. We could refactor so we have these two routes that point to dataset:
> /ds/:name
> /ds/:username/:name

I'm not a fan of this approach, creates another exception case we need to handle. It does make the URL look a little easier for the user, but creates a potential for all sorts of problems when we try to parse this in contrast with other refstring endpoints. I think the `new` stand-in username is a better solution

frontend/src/features/dataset/Dataset.tsx
ramfox: ooooh good idea.

Okay so the reason this doesn't work exactly is we use the fact that `user` is `undefined` to determine some UI behavior. Alternative proposal is to create a `NewUser` const that has everything we need in a `user` to function for our general needs, but is also easy to check if a `user` is a `NewUser` so we can still have the same functionality we had when checking if a `user` is `undefined`

This also formally solidifies that we need to make sure that no one can take the username "new"

This also removes the problem we had before that I was commenting on. Now, we check if the `username` is "new". If so, we replace it with whatever the `User`'s `username` is (which is still "new"  if the user hasn't logged in yet), but won't effect if the `username` is different than the session `user`'s username

yay

PR #116, opened 2021-03-02
test(`SchedulerTestRunner`): create test runner for `scheduler` package tests

ramfox:
And creates a `TestScheduleWorkflowIntegration` test.

The test runner should also shift closer to something like the `NetworkIntegrationTestRunner` in `qri/lib`, as we get a better idea of representing users in qri and qrimatic.

relies on https://github.com/qri-io/dataset/pull/251

There is a bug in dataset that is causing workflows without previous versions to segfault when attempting to `ds.get_body()`

### DIFF
--- a/qrimatic/go.mod
+++ b/qrimatic/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/multiformats/go-multihash v0.0.14
 	github.com/qri-io/apiutil v0.2.0
-	github.com/qri-io/dataset v0.2.1-0.20210128201320-3b1209495e96
+	github.com/qri-io/dataset v0.2.1-0.20210304141850-a4a809d46350
 	github.com/qri-io/ioes v0.1.1
 	github.com/qri-io/iso8601 v0.1.1-0.20201221213213-f31ee4cdc38b
 	github.com/qri-io/qfs v0.5.1-0.20201119141805-fb13393b2d1f
--- a/qrimatic/scheduler/deploy.go
+++ b/qrimatic/scheduler/deploy.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/qri-io/dataset"
 	"github.com/qri-io/qri/base/dsfs"
@@ -33,6 +34,18 @@ func (c *Cron) Deploy(ctx context.Context, inst *lib.Instance, p *DeployParams) 
 	if p.Workflow.DatasetID == "" {
 		return nil, fmt.Errorf("deploy: DatasetID not set")
 	}
+
+	newWorkflow := true
+	if _, err := c.Workflow(ctx, p.Workflow.ID); err == nil {
+		newWorkflow = false
+	}
+
+	now := time.Now()
+	if newWorkflow {
+		p.Workflow.Created = &now
+	}
+	p.Workflow.LatestStart = &now
+
 	dsm := lib.NewDatasetMethods(inst)
 	saveP := &lib.SaveParams{
 		Ref: p.Workflow.DatasetID, // currently the DatasetID is the Ref
@@ -53,13 +66,19 @@ func (c *Cron) Deploy(ctx context.Context, inst *lib.Instance, p *DeployParams) 
 		}
 	}
 
-	ref := &dsref.Ref{
-		Username: res.Peername,
-		Name:     res.Name,
-	}
-	p.Workflow.Complete(ref, inst.Config().Profile.ID)
+	now = NowFunc()
+	p.Workflow.LatestEnd = &now
+	p.Workflow.RunCount++
+	p.Workflow.Status = StatusSucceeded
 
-	// save workflow
+	if newWorkflow {
+		ref := &dsref.Ref{
+			Username: res.Peername,
+			Name:     res.Name,
+		}
+		p.Workflow.Complete(ref, inst.Config().Profile.ID)
+	}
+
 	err = c.Schedule(ctx, p.Workflow)
 	if err != nil {
 		log.Errorw("deploy scheduling", "error", err)
--- a/qrimatic/scheduler/integration_test.go
+++ b/qrimatic/scheduler/integration_test.go
@@ -7,9 +7,12 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/qri-io/dataset"
 	"github.com/qri-io/ioes"
+	"github.com/qri-io/iso8601"
 	"github.com/qri-io/qri/lib"
 	"github.com/qri-io/qri/repo/gen"
 	repotest "github.com/qri-io/qri/repo/test"
@@ -28,7 +31,7 @@ func TestScheduleWorkflowIntegration(t *testing.T) {
 	// TODO (ramfox): until we get multi tenancy the only "username" that
 	// qri will expect is the username of the repo
 	datasetID := fmt.Sprintf("%s/dataset_name", username)
-	w, err := NewCronWorkflow(workflowName, ownerID, datasetID, "R/PT1H")
+	w, err := NewCronWorkflow(workflowName, ownerID, datasetID, "R/PT1S")
 	if err != nil {
 		t.Fatalf("creating workflow: %s", err)
 	}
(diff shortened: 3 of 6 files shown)


qrimatic/scheduler/deploy.go
```diff
@@ -33,6 +34,18 @@ func (c *Cron) Deploy(ctx context.Context, inst *lib.Instance, p *DeployParams)
...
+	if _, err := c.Workflow(ctx, p.Workflow.ID); err == nil {
+		newWorkflow = false
+	}
+
+	now := time.Now()
+	if newWorkflow {
+		p.Workflow.Created = &now
+	}
+	p.Workflow.LatestStart = &now
+
```
ramfox, line 48: okay so this and the below changes are meant to compensate for the fact that `Deploy` was not previously catching any information on the initial "run" of the dataset. This is faking things a bit, and it would be much better in the long run to just route this through `RunWorkflow`. However, `RunWorkflow` just uses the `factory` runner func, which currently assumes that we are running a transform that already exists on the dataset. We need `Deploy` to actually get that transform attached to a dataset.

This also compensates for the fact that since this is currently the only way for qrimatic to attach a transform to a dataset, this is our only path for "updating" a dataset via qrimatic. In those cases, we already have a workflow & we need to make sure that we aren't overwriting anything important when we `Deploy`.

We should probably rethink/remodel the way we want to attach/update transforms & maybe split out the concerns.

It is also a shame that we are unable to use `RunWorkflow`, since it very nicely emits all those lovely events for us & updates the workflow status. However, currently, our code doesn't like it when we have a workflow without an existing dataset, so we have our second reason we can't use `RunWorkflow` in this case (although we will probably want to be in a world where a workflow doesn't have a dataset yet, if necessary).

qrimatic/scheduler/workflow.go
```diff
@@ -117,8 +117,6 @@ func (workflow *Workflow) Complete(ds *dsref.Ref, ownerID string) error {
 	workflow.Name = ds.Human()
 	//TODO (arqu): expand this as this version info is very shallow
 	workflow.VersionInfo = ds.VersionInfo()
-	now := time.Now()
-	workflow.Created = &now
```
ramfox, line 121: this was causing `Workflow.Created` times to be after the dataset has already run. To the user, it makes more sense that workflow was created at the same time the transform was run for the first time, not after.

qrimatic/go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/multiformats/go-multihash v0.0.14
 	github.com/qri-io/apiutil v0.2.0
-	github.com/qri-io/dataset v0.2.1-0.20210128201320-3b1209495e96
+	github.com/qri-io/dataset v0.2.1-0.20210304141850-a4a809d46350
```
b5, line 16: the Dataset PR is merged now. We should update to latest on both qri-io/qri with a new PR and in this PR

qrimatic/scheduler/integration_test.go
```diff
@@ -0,0 +1,290 @@
...
+	"github.com/qri-io/qri/repo/gen"
+	repotest "github.com/qri-io/qri/repo/test"
+	"github.com/qri-io/qri/transform"
+)
+
+func TestScheduleWorkflowIntegration(t *testing.T) {
+
+	username := "integration_test"
+	tr := NewSchedulerTestRunner(t, username)
+	defer tr.Cleanup()
```
b5: I prefer the cleanup as a return value on the TestRunner constructor, that way test authors are forced to adknowledge the need for cleanup:

```go
tr, cleanup := NewSchedulerTestRunner(t, username)
defer cleanup()
```

qrimatic/scheduler/integration_test.go
```diff
@@ -0,0 +1,290 @@
...
+				},
+			},
+		},
+	}
+	ctx := context.Background()
+	// Deploy saves a version of the dataset & schedules the workflow
+	_, err = tr.cron.Deploy(ctx, tr.inst, dp)
+	if err != nil {
+		t.Fatalf("deploying workflow: %s", err)
+	}
```
b5:
```suggestion
	if _, err = tr.cron.Deploy(ctx, tr.inst, dp); err != nil {
		t.Fatalf("deploying workflow: %s", err)
	}
```

qrimatic/scheduler/integration_test.go
```diff
@@ -0,0 +1,290 @@
...
+
+	// ensure workflow is in the cron store
+	deployedWorkflow, err := tr.cron.Workflow(ctx, w.ID)
+	if err != nil {
+		t.Fatalf("getting workflow from cron: %s", err)
+	}
+	if diff := cmp.Diff(w.ID, deployedWorkflow.ID); diff != "" {
+		t.Errorf("deployed workflow ID mismatch (-want +got):\n%s", diff)
+	}
```
b5: this is just comparing strings, replace with:

```go
if w.ID != deployedWorkflow.ID {
```

qrimatic/scheduler/integration_test.go
```diff
@@ -0,0 +1,290 @@
...
+		dsm := lib.NewDatasetMethods(inst)
+
+		return func(ctx context.Context, streams ioes.IOStreams, workflow *Workflow) error {
+			runID := transform.NewRunID()
+
+			// runState = run.NewState(runID)
+			// m.inst.bus.SubscribeID(func(ctx context.Context, e event.Event) error {
+			// 	runState.AddTransformEvent(e)
+			// 	return nil
+			// }, runID)
```
b5: ded code

qrimatic/scheduler/integration_test.go
```diff
@@ -0,0 +1,290 @@
...
+		t.Fatal(err)
+	}
+	tr.repo = &repo
+
+	tr.inst, err = lib.NewInstance(ctx, tr.repo.QriPath, lib.OptIOStreams(ioes.NewDiscardIOStreams()))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	tr.storePath, err = ioutil.TempDir(os.TempDir(), fmt.Sprintf("%s_scheduler_test_runner_store", prefix))
```
b5: no need to pass `os.TempDir()` to `ioutil.TempDir`, the empty string defaults to `os.TempDir()`

qrimatic/scheduler/integration_test.go
```diff
@@ -0,0 +1,290 @@
...
+	tr.cancel()
+	if tr.storePath != "" {
+		os.RemoveAll(tr.storePath)
+	}
+	if tr.repo != nil {
+		tr.repo.Delete()
+	}
+}
+
+func NewSchedulerTestRunner(t *testing.T, prefix string) *SchedulerTestRunner {
```
b5: add `t.Helper()` as the first statement in this function. If it fails it'll keep the line numbers in the test from being all wonky

qrimatic/scheduler/integration_test.go
```diff
@@ -0,0 +1,290 @@
...
+	m := lib.NewDatasetMethods(tr.inst)
+	getParams := &lib.GetParams{
+		Refstr:   datasetID,
+		Selector: "body",
+		All:      true,
+	}
+	getRes, err := m.Get(ctx, getParams)
+	if diff := cmp.Diff("[[1,2,3]]", string(getRes.Bytes)); diff != "" {
+		t.Errorf("deployed dataset body mismatch (-want +got):\n%s", diff)
+	}
```
b5: There's a nice test abstraction to be written here
```go
func AssertBodyEquals(t *testing.T, inst *lib.Instance, refstr, expectBody string) {
  t.Helper()
  m := lib.NewDatasetMethods(tr.inst)
  getParams := &lib.GetParams{
    Refstr:   refstr,
    Selector: "body",
    All:      true,
  }
  getRes, err := m.Get(context.Background(), getParams)
  if diff := cmp.Diff(expectBody, string(getRes.Bytes)); diff != "" {
    t.Errorf("dataset body mismatch (-want +got):\n%s", diff)
   }
}
```

then use this wherever you have `// fetch the dataset and compare bodies`

qrimatic/scheduler/deploy.go
b5: > We should probably rethink/remodel the way we want to attach/update transforms & maybe split out the concerns.

agreed. I think you make a great point that deploying a new dataset should probably require a save-without-apply that works first so we have an initID, and know that the dataset will save, followed by a run, possibly right away (when the user checks the "run script on deploy" option).

The problem there: most transform scripts set the dataset body, and we currently can't save a dataset without a body ☹️

ramfox: > most transform scripts set the dataset body, and we currently can't save a dataset without a body

oof yea. Interesting. Would it be totally gross to... force a body? like save it with an empty `[[]]`? That would force a structure/schema that we would need to make sure gets overwritten. I don't love this, but wondering the time and planning cost/benefit of programming around this vs making qri have to accept a dataset w/o a body

qrimatic/scheduler/integration_test.go
ramfox: LOL

qrimatic/scheduler/integration_test.go
ramfox: LOVE this

b5 (approved):

PR #118, opened 2021-03-08
refactor(backend): delete lots of old, dead things

b5:
rearrange packages, removing the `update` package in the process. I'm not a fan of how this thickens cmd/serve.go, but I think it's worth it to drop a level of indrection by removing the update package & constructing an api service directly.


b5:
@ramfox this is deleting so many things around at once I don't think it's worth a review. I'm going to merge, we'll suffer the consequence 😄

PR #128, opened 2021-03-15
refactor(frontend): simplify workflow editor

b5:
Screen Shot 2021-03-16 at 8 32 10 AM

* redirect to collection instead of dashboard
* strip a number of elements away from workflow editor
* switch workflow editor to fixed-width
* add a "collapse" button for outline view
* move deploy workflow component down blow completion hooks


chriswhong (approved):

chriswhong:
My only feedback is that the collapse button appears to be a back button, but it seems like these are all fixes to make demos run more smoothly now so LGTM.  We will have lots of design improvements coming shortly.

PR #140, opened 2021-04-12
feat: refactor style for dataset nav

chriswhong:
Re-style dataset nav, navbar, and content area to match our new mockups

image


frontend/src/chrome/Button.tsx
```diff
@@ -52,7 +52,7 @@ const Button: React.FC<ButtonProps> = ({
   return (
     <button
       type="button"
-      className={`flex items-center inline-flex justify-center rounded-sm shadow-sm bg-white font-medium focus:outline-none focus:ring focus:ring-offset ring-offset-transparent mt-0 transition-all duration-100 ${colorClasses} ${sizeClasses} ${className}`}
+      className={`flex items-center inline-flex justify-center rounded-md shadow-sm bg-white font-medium focus:outline-none focus:ring focus:ring-offset ring-offset-transparent mt-0 transition-all duration-100 ${colorClasses} ${sizeClasses} ${className}`}
```
b5: we should be replacing string templating for the `className` prop with a call to `classNames`. This whole component can be refactored, removing the need for variables & `switch` statements.

frontend/src/chrome/Icon.tsx
```diff
@@ -152,6 +149,103 @@ const Icon: React.FunctionComponent<IconProps> = ({
     )
   }
 
+  if (icon === 'dashboard') {
```
b5: each of these should get it's own file in a new `icon` directory within the `src/chrome`. That way as icons are changed / added / removed it'll show up clear-as-day in version control.

frontend/src/features/app/App.css
```diff
@@ -1,3 +1,5 @@
+@import url('https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap');
```
b5: This can break qri working in offline mode, and gives our google overlords more analytics data. considering we're using web pack it shouldn't be too hard to bundle `Robato` with the webapp itself. Please remove in the meantime.

frontend/src/chrome/Button.tsx
chriswhong: Looks like you fixed this already

frontend/src/features/app/App.css
chriswhong: Added the files for Roboto to the codebase so they can be bundled.

frontend/src/chrome/Icon.tsx
chriswhong: Added `CustomIcon` which takes an `icon` prop.  `Icon` will return `CustomIcon` if its `icon` prop is not one of the fontawesome icons we are using in the app.

Each custom icon is a component with a react fragment for the inner elements of the SVG, `CustomIcon` contains the actual `svg` element.

frontend/src/chrome/icon/ActivityFeed.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+import CustomIcon, { CustomIconProps } from '../CustomIcon'
+
+const ActivityFeed: React.FC<CustomIconProps> = ({ className }) => (
+  <CustomIcon className={className}>
```
b5: very close, but we should use the {...props} spread operator to avoid having to change these any time `CustomIconProps` changes

```tsx
const ActivityFeed: React.FC<CustomIconProps> = (props) => (
   <CustomIcon {...props}>
```

chriswhong: done

b5 (approved):

PR #141, opened 2021-04-12
feat(frontend): websocket connection healing

b5:
I ended up re-writing this four times, each with a slightly different apprach.

The first factoring had all websocket state in the app reducer. Having an "app" reducer very similar to global state, and made it tempting to call the connection a "backend connection" instead of a "websocket connection".

On the second factoring I moved all websocket-related code into the websocket feature folder, and wrote a set of actions that used a "data" payload to manipulate state in a reducer. All actions were of type "WS_CONNECTION_CHANGE", and the change info was part of the data. I wasn't a fan of having the actions themselves dictate business logic like the number of retries to attempt, etc.

Third factoring moved all of this logic into the reducer, and thinned the actions to just a set of event constants without any payload ("WS_CONNECTED", "WS_CONNECT_FAIL", etc), and the reducer owned how to update state according
to each action type. My plan was to have the websocket middleware fire these actions, and keep state management in the reducer. But, the way redux middlware integrates with reducers makes it difficult/impossible to fire actions & listen for those actions within the same middlware (which, in hindsight, makes sense), which meant the websocket middlware couldn't make use of state changes in the websocket reducer.

So on the fourth go-round I ended up making the websocket middlware the owner of websocket connection state, and used the approach from the second factoring to store a copy of that state in the reducer. In this arrangement redux handles a reducer that UI can listen to, and middlware doesn't have the additional burden of state coordination with an external reducer.

### DIFF
--- a/frontend/src/chrome/StatusDot.tsx
+++ b/frontend/src/chrome/StatusDot.tsx
@@ -10,11 +10,10 @@ export interface StatusDotProps {
   showNoChanges?: boolean
 }
 
-export const StatusDot: React.FunctionComponent<StatusDotProps> = (props) => {
-  const {
-    status,
-    showNoChanges = false
-  } = props
+export const StatusDot: React.FC<StatusDotProps> = ({
+  status,
+  showNoChanges = false
+}) => {
   let statusTooltip
   let statusColor = ''
   switch (status) {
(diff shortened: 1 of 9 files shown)


b5:
@Arqu I'd like to get this into master ASAP, so I've done a solid self review. but feel free to follow up post-merge & I can make changes 😄

PR #142, opened 2021-05-04
feat: layout dataset header

chriswhong:
Home___Qri_Cloud

- Lays out the dataset header with placeholder data attributes
- Adds hover color to dataset nav items
- improves some global nav styling


frontend/src/chrome/CustomIcon.tsx
```diff
@@ -4,14 +4,40 @@ import classNames from 'classnames'
 // TODO(chriswhong): add size prop for sizes other than 24x24
```
ramfox: Can get delete this comment now!

ramfox (approved):
Look at those custom icons!

lgtm, found one stray comment, other than that looks good!

PR #159, opened 2021-05-20
feat: design tweaks for login and signup modals

chriswhong:
Home___Qri_Cloud

Makes login and signup look like the mockups


b5 (approved):
LGTM

PR #161, opened 2021-05-24
feat: style SessionUserMenu

chriswhong:
- styles the sign up and log in buttons for anonymous users
- styles the user menu icon for logged in users
- updates the user menu items
- changes navbar link styles
- adds a default size to `QriLogo` which was not displaying in the navbar due to a missing `size prop`
- adds hover styles to `ButtonGroup` (top my datasets and dashboard buttons)
- animates the caret that is used to expand the dataset nav

Closes #147

PR #165, opened 2021-05-25
Wire up Signup and Login

chriswhong:
Merge #161 first.

- Wires up login and signup actions and reducers
- Uses `store.subscribe` to keep localstorage in sync with `store.session`, so user data and token persist
- Handles errors, shows the user an error in each respective modal.
- Does not handle token refresh, this will be implemented later

### DIFF
--- a/frontend/src/store/api.ts
+++ b/frontend/src/store/api.ts
@@ -25,13 +25,18 @@ export interface ApiAction extends AnyAction {
     segments?: QriRef
     // query is an object of query parameters to be appended to the API call URL
     query?: ApiQuery
-    // body is the body for POST requests
+    // body is the JSON body for POST requests
     body?: object|[]
+    // form is the form data for POST requests (overrides body)
+    // deprecated: form will be removed, don't add callers that populate form
+    form?: object
     // pageInfo is the pagination information
     pageInfo?: ApiPagination
     // map is a function
     // map defaults to the identity function
     map?: (data: any) => any
+    // token is a json web token.  It is normally read from state, but can also be passed in
+    token?: string
   }
 }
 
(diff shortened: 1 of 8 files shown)


frontend/src/store/store.ts
```diff
@@ -74,6 +91,11 @@ export function configureStore(preloadedState?: any) {
     ),
   )
 
+  // automatically sync store.session.token with localstorage
+  store.subscribe(() => {
+    setAuthState(store.getState())
+  });
```
b5, line 100: hmmmmmm, I'm wondering if we can `useSelector` here to avoid passing the entire stateTree to `setAuthState`. I'm pretty sure selectors need to be within a react component (or at least, the child of a `Provider` instance). Would you mind trying to use a selector here & seeing if it explodes?

chriswhong: Can't use `useSelector` outside of a react component, I tried that first.  I can research to see if there is a better way to do this.

b5: nah, it's fine. Let's just keep an eye on compiled app performance. If it degrades we'll come back here & see if it's causing problems

frontend/src/store/api.ts
```diff
@@ -25,13 +25,17 @@ export interface ApiAction extends AnyAction {
     segments?: QriRef
     // query is an object of query parameters to be appended to the API call URL
     query?: ApiQuery
-    // body is the body for POST requests
+    // body is the JSON body for POST requests
     body?: object|[]
+    // form is the form data for POST requests (overrides body)
+    form?: object
```
b5: add a deprecation notice:
```tsx
// form is the form data for POST requests. overrides body
// deprecated: form will be removed, don't add callers that populate form
form?: object
```

frontend/src/store/api.ts
```diff
@@ -8,7 +8,7 @@ import { QriRef } from '../qri/ref'
 // CALL_API is a global, unique constant for passing actions to API middleware
 export const CALL_API = Symbol('CALL_API')
 
-export const API_BASE_URL = process.env.REACT_APP_API_BASE_URL || 'http://localhost:2503'
+export const API_BASE_URL = process.env.REACT_APP_API_BASE_URL || 'http://localhost:3002'
```
b5: please revert to port `2503`

b5 (approved):
looks great! one or two nits then merge city

PR #166, opened 2021-05-27
Previews, Navbar Search, readmes

chriswhong:
- hydrates dataset previews
- enables navbar search (hides navbar search when on `/search`)
- ensures that a search term passed in from an external route via query params is populated in the `/search` text input
- removes an unused API call for paginated body (to be used later)
- refactors the API call for a rendered body.

### DIFF
--- a/frontend/src/features/dsComponents/body/Body.tsx
+++ b/frontend/src/features/dsComponents/body/Body.tsx
@@ -1,26 +1,26 @@
 import React, { useEffect } from 'react'
+import { useDispatch } from 'react-redux'
 // import { Action } from 'redux'
 
+import Dataset,  { Structure, schemaToColumns, ColumnProperties } from '../../../qri/dataset'
+import BodyTable from './BodyTable'
+import BodyJson from './BodyJson'
+import { loadBody } from '../../dataset/state/datasetActions'
+import { newQriRef } from '../../../qri/ref'
 // import { ApiActionThunk } from '../../../store/api'
 // import { DetailsType, StatsDetails, Details } from '../../../models/details'
-import Dataset,  { Structure, schemaToColumns, ColumnProperties } from '../../../qri/dataset'
 // import { RouteProps } from '../../../models/store'
 // import { fetchBody, fetchCommitBody } from '../../../actions/api'
 // import { setDetailsBar } from '../../../actions/ui'
 // import { selectDataset, selectWorkingDataset, selectDatasetStats, selectWorkingStats, selectDetails, selectDatasetBodyPageInfo, selectWorkingDatasetBodyPageInfo, selectWorkingStatusInfo } from '../../../selections'
 // import { QriRef, qriRefFromRoute } from '../../../models/qriRef'
-
-import BodyTable from './BodyTable'
-import BodyJson from './BodyJson'
-import { useDispatch } from 'react-redux'
-import { loadBody } from '../../dataset/state/datasetActions'
-import { newQriRef } from '../../../qri/ref'
 // import ParseError from '../ParseError'
 // import hasParseError from '../../../utils/hasParseError'
 // import { connectComponentToProps } from '../../../utils/connectComponentToProps'
 
 export interface BodyProps {
   data: Dataset
+  preview?: boolean
   // stats: IStatTypes[]
   // details: Details
   // pageInfo: PageInfo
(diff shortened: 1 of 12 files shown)


frontend/src/features/search/BigSearchBox.tsx
```diff
@@ -21,6 +21,10 @@ const BigSearchBox: React.FC<BigSearchBoxProps> = ({ onSubmit, value='', classNa
     onSubmit(stateValue)
   }
 
+  React.useEffect(() => {
+    setStateValue(value)
+  }, [value])
```
b5, line 29: I'd love a comment clarifying that this updates state if the value prop changes in parent components. It feels a _little_ smelly to me, but not enough to worry about. Adding a comment helps say "yes, this is a bit smelly" to future readers.

frontend/src/features/dsComponents/body/Body.tsx
```diff
@@ -52,14 +52,15 @@ const extractColumnHeaders = (structure: Structure, value: any[]): ColumnPropert
 const Body: React.FunctionComponent<BodyProps> = ({
```
b5: React.FC

frontend/src/features/dsComponents/body/Body.tsx
```diff
@@ -52,14 +52,15 @@ const extractColumnHeaders = (structure: Structure, value: any[]): ColumnPropert
...
+  // const { path, name, peername: username } = data
 
   // list out dependencies on dataset body individually for proper memoization
-  useEffect(() => {
-    dispatch(loadBody(newQriRef({ path, name, username }), 1, 100))
-  }, [dispatch, path, name, username])
+  // TODO(chriswhong): reinstate when we want full body UI
+  // useEffect(() => {
+    // dispatch(loadBody(newQriRef({ path, name, username }), 1, 100))
+  // }, [dispatch, path, name, username])
```
b5: Instead of commenting this out, add a `preview?: boolean` to `BodyProps`, default it to false & pass `preview` to this effect:

```tsx
useEffect(() => {
  if (preview) { return }
  dispatch(loadBody(newQriRef({ path, name, username }), 1, 100))
}), [dispatch, path, name, username, preview])
```

generally, the `preview` prop should be used to flip this component's behaviour

frontend/src/features/dsPreview/state/dsPreviewState.ts
```diff
@@ -0,0 +1,31 @@
+import { createReducer } from '@reduxjs/toolkit'
+
+import { RootState } from '../../../store/store';
+
+export const selectDsPreview = (state: RootState): any => state.dsPreview.preview
+
+export const selectIsDsPreviewLoading = (state: RootState): boolean => state.dsPreview.loading
+
+export interface DsPreviewState {
+  preview: any
```
b5: we need a real type here. `Dataset` should work. Also, should be fine to make this property optional

frontend/src/features/dsPreview/state/dsPreviewState.ts
```diff
@@ -0,0 +1,31 @@
...
+  loading: boolean
+}
+
+const initialState: DsPreviewState = {
+  preview: {},
```
b5: drop this from initial state

frontend/src/features/dsPreview/state/dsPreviewState.ts
```diff
@@ -0,0 +1,31 @@
...
+  'API_DATASETPREVIEW_REQUEST': (state) => {
+    state.preview = {}
+    state.loading = true
+  },
+  'API_DATASETPREVIEW_SUCCESS': (state, action) => {
+    state.preview = action.payload.data
+    state.loading = false
+  },
+  'API_DATASETPREVIEW_FAILURE': (state) => {
+    state.preview = false
```
b5: set preview to `undefined`, and don't forget to set loading to false

frontend/src/features/dsPreview/state/dsPreviewState.ts
```diff
@@ -0,0 +1,31 @@
...
+export const selectIsDsPreviewLoading = (state: RootState): boolean => state.dsPreview.loading
+
+export interface DsPreviewState {
+  preview: any
+  loading: boolean
+}
+
+const initialState: DsPreviewState = {
+  preview: {},
+  loading: true
```
b5: this should be `false`

b5 (approved):
Wahoo!

PR #184, opened 2021-06-03
feat: add user profile page, improve search

chriswhong:
Adds user profile routes with two dataset lists:
- user datasets `/:username`
- datasets the user is following `/:username/following`

Both lists share query params of type `UserProfileParams`.  Introduces a new pattern for query params:

1) on mount or update, read query params from URL
2) convert params from the URL into valid `UserProfileParams` by adding default values
3) fetch data using `UserProfileParams`
4) when user interacts with pagination, query, or sort, etc, modify the `UserProfileParams`
5) update the URL with a "cleaned" version of `UserProfileParams`, (excluding default values)
6) go to 1

Updates Search to use the same pattern for query params.

TODO:
- improve loading states
- add a sticky header on scroll
- handle no results (e.g. user doesn't follow any datasets)

### DIFF
--- a/frontend/src/chrome/ExternalLink.tsx
+++ b/frontend/src/chrome/ExternalLink.tsx
@@ -4,6 +4,7 @@ export interface ExternalLinkProps {
   id?: string
   to: string
   children: React.ReactNode
+  // providing className prop will override the component's color classes
   className?: string
   tooltip?: string
 }
@@ -22,7 +23,7 @@ export const ExternalLink: React.FunctionComponent<ExternalLinkProps> = ({
     href={to}
     target='_blank'
     rel="noopener noreferrer"
-    className={`text-qriblue hover:text-qriblue-600 hover:cursor-pointer ${className}`}
+    className={`${className || 'text-qriblue hover:text-qriblue-600'} hover:cursor-pointer`}
     data-tip={tooltip}
   >
     {children}
--- a/frontend/src/features/search/Search.tsx
+++ b/frontend/src/features/search/Search.tsx
@@ -7,10 +7,14 @@ import { loadSearchResults } from './state/searchActions'
 import { selectSearchResults, selectSearchPageInfo } from './state/searchState'
 import NavBar from '../navbar/NavBar';
 import BigSearchBox from './BigSearchBox';
-import SearchResultItem from './SearchResultItem'
 import PageControl, { PageChangeObject } from '../../chrome/PageControl'
 import Footer from '../footer/Footer'
 import { NewSearchParams } from '../../qri/search'
+import DatasetList from '../../chrome/DatasetList'
+import ContentBox from '../../chrome/ContentBox'
+import Icon from '../../chrome/Icon'
+import DropdownMenu from '../../chrome/DropdownMenu'
+import { CleanSearchParams } from '../../qri/search'
 
 const Search: React.FC<{}> = () => {
   const dispatch = useDispatch()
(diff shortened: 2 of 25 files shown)


frontend/src/features/search/SearchResultItem.tsx
```diff
@@ -22,7 +22,13 @@ const SearchResultItem: React.FC<SearchResultItemProps> = ({ dataset }) => {
...
+    let title, description
+
+    if (meta) {
+      title = meta.title
+      description = meta.description
+    }
+
     if (title === undefined) {
       title = datasetReference
     }
```
ramfox: you can use conditional chaining/ternary operator to simplify this:

```
const title = meta ? meta.title : datasetReference 
const description = meta?.description
```

the above is for if you want to show the `datasetReference` when the meta is undefined
If you want to show the `datasetReference` if the meta.title is undefined, null, or an empty string:

```
const title = meta?.title || datasetReference
```

frontend/src/chrome/ExternalLink.tsx
```diff
@@ -22,7 +22,7 @@ export const ExternalLink: React.FunctionComponent<ExternalLinkProps> = ({
     href={to}
     target='_blank'
     rel="noopener noreferrer"
-    className={`text-qriblue hover:text-qriblue-600 hover:cursor-pointer ${className}`}
+    className={`${className || 'text-qriblue hover:text-qriblue-600 hover:cursor-pointer'} hover:cursor-pointer`}
```
ramfox: might want to add a comment to the `ExternalLinkProps` className field saying that adding a `className` overrides all default styling.

frontend/src/features/search/state/searchActions.ts
```diff
@@ -1,30 +1,36 @@
...
 }
 
-function fetchSearchResults (q: string, page: number, pageSize: number): ApiAction {
+const mapFrontendParams = (frontendParams: SearchParams) => {
+  // map frontend 'sort' param to backend 'orderBy'
+  return {
+    q: frontendParams.q,
+    orderBy: frontendParams.sort === 'recentlyupdated' ? 'created,desc' : 'name,asc'
+  }
+}
```
ramfox, line 21: good isolation, clues for what needs to be reconciled between both systems.

frontend/src/features/userProfile/UserProfileHeader.tsx
```diff
@@ -0,0 +1,47 @@
...
+        <div className='text-qrinavy text-sm'>{username}</div>
+        <div className='text-qrigray-400 text-sm'>{description || 'I\'m a Qri user'}</div>
+      </div>
+      <div className='flex flex-shrink-0 ml-4 flex-col justify-end items-end'>
+        <div className='flex justify-end mb-2'>
+          <IconLink icon='github' link='https://github.com/chriswhong' />
+          <IconLink icon='twitter' link='https://twitter.com/chriswhong' />
+          <IconLink icon='globe' link='https://chriswhong.com' />
+        </div>
+        <div className='text-xs text-qrigray-400'>Qri user since 2020</div>
```
ramfox: is this meant to be hardcoded for now?
We do have a `profile.created` field we can use with `moment`, no? and if it's empty or equal to zero, we can add a default, or not show it.

frontend/src/features/userProfile/state/userProfileActions.ts
```diff
@@ -0,0 +1,78 @@
...
+const mapSearchResults = (results: any[]) => {
+  return results.map((d) => NewSearchResult(d))
+}
+
+const mapFrontendParams = (frontendParams: UserProfileParams) => {
+  // map frontend 'sort' param to backend 'orderBy'
+  return {
+    orderBy: frontendParams.sort === 'recentlyupdated' ? 'created,desc' : 'name,asc'
+  }
+}
```
ramfox: just a thought, maybe moving the mapping functions from the `actions` file and `reducer` file to a separate `map` file & we know that we are done syncing between the two systems when we can remove the functions that aren't just a one to one conversion to the given interface?

frontend/src/features/userProfile/state/userProfileActions.ts
```diff
@@ -0,0 +1,78 @@
...
+      map: mapSearchResults
+    }
+  }
+}
+
+function fetchUserProfileFollowing (username: string, userProfileParams: UserProfileParams): ApiAction {
+  return {
+    type: 'userprofilefollowing',
+    [CALL_API]: {
+      endpoint: `follow/${username}`,
```
ramfox, line 68: probs more a note for backend & take it or leave it, but looking at this, I would expect an endpoint called `follow` to be how I follow a specific user, not showing me who that user follows.

frontend/src/features/userProfile/state/userProfileActions.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export function loadUserProfile(username: string): ApiActionThunk {
+  return async (dispatch, getState) => {
+    return dispatch(fetchUserProfile(username))
+  }
+}
+
+export function loadUserProfileDatasets(username: string, userProfileParams: UserProfileParams): ApiActionThunk {
+  return async (dispatch, getState) => {
+    return dispatch(fetchUserProfileDatasets(username, userProfileParams))
```
ramfox, line 16: The name `UserProfileParams` implies that the results of the fetch would be user profiles. Do we have a `DatasetListParams`, or a more general `ListParams` that can be used?

frontend/src/features/userProfile/UserProfileDatasetList.tsx
```diff
@@ -0,0 +1,84 @@
...
+              Sort By
+              <Icon icon='caretDown' size='2xs' className='ml-3' />
+            </div>
+          </DropdownMenu>
+        </div>
+        <DatasetList datasets={results} />
+      </ContentBox>
+      <PageControl
+        pageInfo={pageInfo}
+        queryParams={CleanUserProfileParams(userProfileParams)}
```
ramfox: the only thing I'm confused about is why we are pulling out default values from the userProfileParams

frontend/src/features/userProfile/state/userProfileState.ts
```diff
@@ -0,0 +1,94 @@
...
+  'API_USERPROFILEFOLLOWING_REQUEST': (state: UserProfileState, action) => {
+    state.following.loading = true
+  },
+  'API_USERPROFILEFOLLOWING_SUCCESS': (state: UserProfileState, action) => {
+    state.following.results = action.payload.data
+    state.following.pageInfo = {
+      // TODO(chriswhong): this API doesn't return resultCount,
+      // this will ensure it exists in the frontend in development
+      ...action.payload.pagination,
+      resultCount: 29
```
ramfox: what about using the `state.following.results` to calculate the resultCount?
```
resultCount = state.following?.results : state.following.results.length : 0
```

frontend/src/features/search/Search.tsx
```diff
@@ -23,57 +27,104 @@ const Search: React.FC<{}> = () => {
...
   const handlePageChange = ({ selected: pageIndex }: PageChangeObject) => {
-    const newPage = pageIndex + 1
-    const newParams = new URLSearchParams(`q=${q}&page=${newPage}`)
-    history.push({ search: newParams.toString() })
-    if (scrollContainer.current) {
+    updateQueryParams({ page: pageIndex + 1 })
+  }
+
+  // merges new query params with existing params, updates history
+  const updateQueryParams = (newQueryParams: Record<string, any>) => {
```
ramfox, line 49: dig it! like that there is only one place to look to see by what rules the params are shifting

frontend/src/features/userProfile/UserProfileDatasetList.tsx
ramfox: wait i think i get it now, after seeing how this is handled in the `Search` page. This is sanitizing what the users see in the url, not what we use to actually fetch from the backend, correct? So if the requests are the "default" we aren't crowding the url.

frontend/src/features/userProfile/UserProfile.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  // if the query string ever changes, fetch new data
+  useEffect(() => {
+    dispatch(loadUserProfile(usernameParam))
+    dispatch(loadUserProfileDatasets(usernameParam, userProfileParams))
+    dispatch(loadUserProfileFollowing(usernameParam, userProfileParams))
+  }, [usernameParam, page, sort, dispatch])
+
+  // merges new query params with existing params, updates history
+  const updateQueryParams = (newQueryParams: Record<string, any>) => {
```
ramfox, line 56: again, like the pattern of pulling all the mutations into one function, makes for easy reading/grok-ing!

ramfox (approved):
One or two style nits I think should be changed, and some questions about naming of our listing params, but since that is an ongoing conversation anyway, not something that needs to be fixed right now: LGMT

frontend/src/features/userProfile/state/userProfileState.ts
chriswhong: results is just one page of results, so we still don't know the total.  I just set this to 29 so that it forces pagination controls to display, we could also just use results.length and it will force it to be only one page.  Either way we need to have the backend return the `resultCount`

frontend/src/features/userProfile/UserProfileDatasetList.tsx
chriswhong: yep, this just keeps the URL free of default values.

frontend/src/features/userProfile/state/userProfileActions.ts
chriswhong: `UserProfileParams` is meant to say "these are the valid url query params for the `UserProfile` view/route.  For now both lists on this view allow for the same params, but these will probably diverge, so I think we will need to be more specific on these and have a unique set of params for each list (or perhaps extend a generic `ListParams`)

frontend/src/features/userProfile/UserProfile.tsx
chriswhong: :)

frontend/src/features/userProfile/UserProfileHeader.tsx
chriswhong: Good catch.  I think we are using `date-fns` now, so I'll code it up.

frontend/src/features/userProfile/state/userProfileState.ts
chriswhong: I'll leave the TODO in but add the ternary you suggested so that we aren't showing page links that don't exist.

frontend/src/features/userProfile/state/userProfileActions.ts
chriswhong: I think I prefer them nearby where they are used so they're in your face if you visit this file.  If we end up with a lot of them or lots of duplication, we could centralize.

frontend/src/features/userProfile/state/userProfileActions.ts
chriswhong: I'll change this to `UserProfileDatasetListParams`

PR #185, opened 2021-06-04
refactor(frontend): adjust for local API

b5:


frontend/src/features/deploy/state/deployActions.ts
```diff
@@ -3,17 +3,41 @@ import { NewWorkflow, Workflow, workflowInfoFromWorkflow, workflowScriptString }
...
+// }
+
 export function deployWorkflow(w: Workflow): ApiActionThunk {
   return async (dispatch, getState) => {
     return dispatch({
       type: 'deploy',
       
       [CALL_API]: {
-        endpoint: 'deploy',
+        endpoint: '/auto/deploy',
```
b5: possible different cloud endpoint

frontend/src/features/dsPreview/state/dsPreviewActions.ts
```diff
@@ -12,7 +12,7 @@ function fetchDsPreview (ref: QriRef): ApiAction {
     type: 'datasetpreview',
     ref,
     [CALL_API]: {
-      endpoint: 'dataset_summary',
+      endpoint: 'ds/get',
```
b5: cloud & local differ here. In my code spike I've adjusted `qri get` with no selector to return a preview

frontend/src/features/deploy/state/deployActions.ts
Arqu: Cloud doesn't have this one yet. But there is probably a long list of endpoints that generally do need to be brought over from core into cloud too. I'll make an issue for myself to review the entire API spec and filter out the ones that need to be added.

frontend/src/features/dsPreview/state/dsPreviewActions.ts
Arqu: q: Should this be the case? Shouldn't get return the "full" response while previews should be separate. We do also have `/remote/previews` albeit a bit different thing there. I guess using `/ds/get` to fetch previews should be fine once cursors and proper paging are set up. 

Still not sure if we should bend `get` into this. And am more for setting up `/ds/preview` for it for now. The route fix as requested can land as soon as the changes in core for `get` are on master.

b5:
superseded by #206

PR #190, opened 2021-06-08
feat: add sticky mini-headers to preview and profile views

chriswhong:
Adds sticky mini-headers to preview and profile views, which animate in once the user scrolls the main header out of view.
Jun-08-2021 15-07-59


b5 (approved):
LGTM

PR #193, opened 2021-06-11
Various improvements

chriswhong:
Various fixes and tweaks, from a pairing session with Tefy

- [x] Button Hover State: Black Outline Button should show pink.  Blue outline button should get darker on hover
- [x] Add IconButton component that is ALWAYS square (implement for download button)
- [x] Typing additional characters in input in modal jumps to the end
- [x] Preview not full height when loading
- [x] Gray Background on Dropdown Menu Items should be full-width
- [x] Dataset Menu Tooltips render too far to the right
- [x] Add Courier-like webfoot for font-mono
- [x] Sideways scrolling on dataset preview (http://localhost:3000/ds/chriswhong-demo/new_york_city_covid_19_hot_spot_zones/preview)
- [x] Commit message should show ellipsis and not wrap to new Lin
- [x] Tweak padding around README on History Readme Tab
- [x] Improve Loading State for Collection List (shows “No Datasets to Display”)
- [x] Sort Arrows on Collection List do not match design (see if we can override)


frontend/src/chrome/forms/TextInput.tsx
```diff
@@ -12,7 +12,7 @@ interface TextInputProps {
   label?: string
   labelTooltip?: string
   onKeyDown?: (e: React.KeyboardEvent<HTMLInputElement>) => void | undefined
-  onChange?: (e: React.ChangeEvent) => void | undefined
+  onChange?: (value: string) => void | undefined
```
b5, line 15: drop `| undefined`. `onChange` is already optional. Secondly, don't eliminate the `e` event entirely. We need it whenever event propagation needs to be manipulated. Pass the event as an optional second argument instead.

frontend/src/features/app/App.tsx
```diff
@@ -13,9 +13,9 @@ import SnackBar from '../snackBar/SnackBar';
 const App: React.FC<any> = () => {
   const dispatch = useDispatch()
 
-  useEffect(() => {
-    dispatch(wsConnect())
-  })
+  // useEffect(() => {
+  //   dispatch(wsConnect())
+  // })
```
b5: do not commit this change.

frontend/src/features/collection/WorkflowsTable.tsx
```diff
@@ -45,7 +45,7 @@ const fieldValue = (row: VersionInfo, field: string) => {
 // column sort function for react-data-table
```
b5, line 45: we should rename this file to `CollectionTable`

frontend/src/features/collection/state/collectionState.ts
```diff
@@ -24,7 +24,15 @@ export const selectCollection = (state: RootState): VersionInfo[] => {
...
   })
-  return ordered
+  return ordered.sort((a, b) => {
+    const aVal = `${a.username}/${a.name}`
+    const bVal = `${b.username}/${b.name}`
+    if (aVal === bVal) {
+      return 0
+    } else {
+      return (aVal < bVal) ? -1 : 1
+    }
```
b5: string templating in a sort function is expensive. Instead try this:

```ts
return ordered.sort((a,b) => {
  if (a.username === b.username && a.name === b.name) { return 0 }
  else if (a.username < b.username ||  a.name < b.name) { return -1 }
  return 1
})
```

definitely play with this refactored function to make sure it behaves the same

frontend/src/features/dataset/DatasetSideNavItem.tsx
```diff
@@ -18,34 +18,37 @@ const DatasetSideNavItem: React.FC<DatasetSideNavItemProps> = ({ id, icon, label
...
-              right: 10
-            }}
-          >
-            {tooltip}
-          </ReactTooltip>
-        )}
-      </Link>
-    </div>
+    <>
+      <div className='mb-4 inline-block'>
```
b5, line 22: this feels pretty gross. `inline-block` + `<br />` feels just like `display: block`, no?

b5 (changes requested):

frontend/src/features/dataset/DatasetSideNavItem.tsx
chriswhong: When displayed as block, the div fills the entire width of the container and the tooltip appears too far to the right.  This approach makes the tooltips appear in the correct place, just to the right of the text.

frontend/src/chrome/forms/TextInput.tsx
chriswhong: This was causing a bug where the input re-renders, causing the cursor to jump to the end of the input.  This approach fixes the bug, but requires that higher-level components just get the value string, not the full onChange event.

frontend/src/features/dataset/DatasetSideNavItem.tsx
b5: ok, please add a comment clarifying as such

frontend/src/chrome/forms/TextInput.tsx
b5: hrm, ok there's totally something weird going on here, but I'm not confident _only_ changing the argument is a fix. I suspect this strange combination of `useEffect` & `useState` is part of the problem. Let's accept this change so we can keep moving, but let's file an issue titled: `Correct Controlled Component pattern in TextInput`. & like to here: https://reactjs.org/docs/forms.html

chriswhong: Done https://github.com/qri-io/qrimatic/issues/196

b5 (approved):
LGTM

PR #194, opened 2021-06-11
Cw cleanup

chriswhong:
- fixes a bug: when anonymous users load a secure route, they were redirected to `/splash` which does not exist, so was treated as seeking the profile for user `splash`

- Adds default content to `/search` with links to three search terms to try out
- Improves links throughout the site, making use of `TextLink` and `IconLink` in various places to standardize the color, hover styles, etc.  `ExternalLink` no longer applies classes of its own and is used for marking up the `a` tag with target=_blank etc.

### DIFF
--- a/frontend/src/chrome/TextLink.tsx
+++ b/frontend/src/chrome/TextLink.tsx
@@ -1,0 +1,43 @@
+import React from 'react'
+import { Link } from 'react-router-dom'
+import classNames from 'classnames'
+
+import ExternalLink from './ExternalLink'
+
+interface TextLinkProps {
+  to?: string
+  // appends arbitrary classnames to the rendered element
+  className?: string
+  // overrides the default color and hover color
+  colorClassName?: string
+  onClick?: () => void
+}
+
+const defaultColorClassname = 'text-qriblue hover:text-qriblue-800'
+
+const TextLink: React.FC<TextLinkProps> = ({
+  to='',
+  onClick,
+  className,
+  colorClassName = defaultColorClassname,
+  children
+}) => {
+
+  const combinedClassNames = classNames('hover:cursor-pointer transition-all duration-100', colorClassName, className)
+
+  if (to) {
+    if (to.indexOf('http') === 0) {
+      return <ExternalLink to={to} className={combinedClassNames}>{children}</ExternalLink>
+    } else {
+      return <Link to={to} className={combinedClassNames}>{children}</Link>
+    }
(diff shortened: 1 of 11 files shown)


frontend/src/chrome/TextLink.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
+import { Link } from 'react-router-dom'
+
+import ExternalLink from './ExternalLink'
+
+interface TextLinkProps {
+  to?: string
+  onClick?: () => void
```
b5: function props last

frontend/src/chrome/TextLink.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+const TextLink: React.FC<TextLinkProps> = ({
+  to='',
+  onClick,
+  className,
+  colorClassName = defaultColorClassname,
+  children
+}) => {
+
+  const classNames = `${transitionAndCursorClassname} ${colorClassName} ${className}`
```
b5: use `classNames`
```suggestion
  const className = classNames(''hover:cursor-pointer transition-all duration-100'', colorClassName, className)
```

inline the `transitionAndCursorClassname` const

frontend/src/chrome/TextLink.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  if (to) {
+    if (to.includes('http')) {
```
b5: lots of strings can include "http" and not be external links. Use:

```
if (to.indexOf('http') === 0)
```

frontend/src/chrome/TextLink.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
```
b5, line 1: Any reason not to rename this to just `Link`? I think it's a good thing that it collides with `react-router-dom`'s `Link`, because they do the same thing.

b5 (approved):
a few nits, then LGTM

frontend/src/chrome/TextLink.tsx
chriswhong: There are a lot of places where we are using `Link` which should use this instead.  I'll open an issue to audit the use of `Link` and use `TextLink` instead where we can.

chriswhong: https://github.com/qri-io/qrimatic/issues/197

PR #198, opened 2021-06-11
various cleanup items

chriswhong:
- fixes a bug where qriRef was undefined in `DatasetHeader`
- use mono font in profile sticky header
- add a simple prompt when user has no datasets, link to Qri CLI quickstart
- user can hit enter to submit login/signup forms
- hide my datasets/dashboard nav items when user is not logged in
- make readme error text gray
- add some margin above Description in preview
- make usernames clickable in dataset header, links to user's profile
- adjust padding for description in preview page
- clean up table borders!


b5 (approved):
LGTM

PR #199, opened 2021-06-14
refactor(fe): remove backend code and move frontend to root

Arqu:
@chriswhong Can I bother you to pick this PR up and fix up the README's before merging.


chriswhong:
Good to go, I don't think we need any of the create react app default README.

b5 (approved):
Great stuff! Let's do  this now while the PR queue is small. Thanks @Arqu!

PR #202, opened 2021-06-18
cw output states

chriswhong:
- Adds dataset preview to the workflow editor (moves it from appearing as output for the last step into its own top-level spot after the steps)
- General workflow UX improvements, including Dry Run button in the header, status icon improvements, placeholder text, etc.
- Improves and styles output for each step


b5:
superseded by #206

PR #203, opened 2021-06-18
fix(auth): update routes for fetching session user profile

Arqu:
Part of https://github.com/qri-io/qri/issues/1806

### DIFF
--- a/src/features/session/state/sessionActions.ts
+++ b/src/features/session/state/sessionActions.ts
@@ -18,8 +18,8 @@ export function logIn (username: string, password: string): ApiActionThunk {
       const token = tokenResponse.payload.data.access_token
 
       // get user data
-      // pass access_token directly to fetchSession so it can be used immediately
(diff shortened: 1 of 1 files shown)


b5 (commented):
hrm, this seems to change the meaning of this function. I'd say leave this with me @Arqu & I can add a commit that gets this into mergeable shape

Arqu:
Hmmm, so the thing was we didn't have a concept of identity and session on qri core until now. Profile gets us most of the way and it nearly behaves the same now with tokens being used. Only missing piece is that `identity/session` should error if the user is not providing a valid token, this just gives the base profile (which from a usage perspective might not be that wrong?)

b5 (approved):
looks good to me. Your comment points to a need on the qri core side to sort out what is & isn't "good behavior" now that sessions have been implicitly added.

Arqu:
Made an issue out of it.

PR #204, opened 2021-06-21
structure and meta styling

chriswhong:
- Styles structure and meta tabs to match mockups
- Removes `.eslintcache` from version control

This adds `important: true` to `tailwind.css` which makes all tailwind styles `!important`.  This was necessary to override the default styles of `react-tooltip` for a tooltip used in the `LabeledStats` component.  If we don't want to go this route, we'll need to find another solution to make custom-styled tooltips.


tailwind.config.js
```diff
@@ -133,5 +133,6 @@ module.exports = {
   plugins: [
     require('@tailwindcss/forms'),
     require('@tailwindcss/line-clamp'),
-  ]
+  ],
+  important: true
```
b5: I don't think this is a good idea. Can we instead style react-tooltip directly with normal CSS instead of using tailwind?

chriswhong: Will do

chriswhong: Neat, tailwind has an `@apply` that lets you just throw tailwind classes into your normal CSS.

```
.light-tooltip {
  @apply rounded-md shadow-lg bg-white ring-1 ring-black ring-opacity-5 opacity-100 p-4 !important;
}
```

src/features/dsComponents/schema/DynamicEditField.tsx
```diff
@@ -140,7 +140,7 @@ const DynamicEditField: React.FunctionComponent<DynamicEditFieldProps> = ({
   }
 
   return (
-    <div style={{ minWidth }} className={classNames('overflow-hidden whitespace-pre min-h-full overflow-ellipsis', { 'invalid': !isValid, 'dynamic-edit-field-large': large, 'focused': focused, 'dynamic-edit-field-editable': editable })} >
+    <div style={{ minWidth }} className={classNames('whitespace-pre min-h-full', { 'invalid': !isValid, 'dynamic-edit-field-large': large, 'focused': focused, 'dynamic-edit-field-editable': editable })} >
```
b5: would prefer to see this spread across multiple lines

src/features/dsComponents/meta/Meta.tsx
```diff
@@ -28,9 +28,9 @@ const renderValue = (value: string | string[] | object) => {
   }
 }
 
-const renderChips = (value: string[] | undefined) => (
+export const renderChips = (value: string[] | undefined) => (
   <div>
-    {value && value.map((d, i) => (<span key={i} className='text-xs rounded bg-gray-300 px-2 py-1 mr-2'>{d}</span>))}
+    {value && value.map((d, i) => (<span key={i} className='leading-tight text-gray-400 text-xs tracking-wider inline-block border border-qrigray-400 rounded-md px-2 py-1 mr-1 mb-1'>{d}</span>))}
   </div>
 )
```
b5: This should be made into a `Chips` or `MetaChips` component

src/features/dsComponents/structure/FormatConfigHistory.tsx
```diff
@@ -13,11 +13,11 @@ const FormatConfigHistory: React.FunctionComponent<FormatConfigHistoryProps> = (
   const format = structure.format
   const formatConfig = structure.formatConfig
```
b5: rename this component to just `FormatConfig`, add a `showTrueOptionsOnly` prop, & use it when calling from history. This component is super old 😄

src/features/dsComponents/schema/SchemaItem.tsx
```diff
@@ -60,59 +60,75 @@ const SchemaItem: React.FunctionComponent<SchemaItemProps> = ({
...
+              placeholder='validation'
+              value={data.validation || ''}
+              onChange={handleDynamicEditChange}
+              allowEmpty expanded={expanded}
+              minWidth={100}
+              editable={editable}
+            />
+          </div>
+        </td>
+      */}
```
b5: let's remove this entirely

src/features/dsComponents/schema/SchemaItem.tsx
```diff
@@ -60,59 +60,75 @@ const SchemaItem: React.FunctionComponent<SchemaItemProps> = ({
...
-          onChange={handleDynamicEditChange}
-          allowEmpty expanded={expanded}
-          minWidth={100}
-          editable={editable}
-        />
+      {/*
+        <td className={tdCellClassName} onClick={() => setExpanded((prev) => !prev)} >
+          <Icon icon={expanded ? 'caretDown' : 'caretRight'} size='md' color='medium'/>
+        </td>
+      */}
```
b5: remove this for now.

src/features/dsComponents/schema/Schema.tsx
```diff
@@ -32,30 +33,53 @@ const Schema: React.FC<SchemaProps> = ({
...
+                description
+              </div>
+            </th>
+            {/*
+              <th className={thClassName}>
+                <div className={thCellClassName}>
+                  validation
+                </div>
+              </th>
+            */}
```
b5: remove

src/chrome/LabeledStats.tsx
```diff
@@ -30,40 +34,78 @@ export interface LabeledStatsProps {
...
 
+export function getStats (data: Structure): any[] {
+  return [
+    { 'label': 'format', 'value': data.format ? data.format.toUpperCase() : 'unknown' },
+    { 'label': 'body size', 'value': data.length ? fileSize(data.length) : '—' },
+    { 'label': 'entries', 'value': abbreviateNumber(data.entries) || '—' },
+    { 'label': 'errors', 'value': data.errCount ? abbreviateNumber(data.errCount) : '—' },
+    { 'label': 'depth', 'value': data.depth || '—' }
+  ]
+}
```
b5: the stats prop is intended to be generic. Move this into `/src/qri/dataset.ts`, somewhere near the declaration of the structure component, restore the `Stat` interface, and pass the result of calling `getStats` to the component

chriswhong: LabeledStats is rendering a tooltip that contains info from `structure`, so we need to pass structure in as a prop and call `getProps()` from within this component.  Either that or we need a separate component to render the tooltip that will select structure on its own, which feels wrong.

I've moved the `Stat` interface to `/qri/dataset.ts` and made sure that `getStats()` is returning `Stat[]`

PR #207, opened 2021-06-23
refactor(api): aligning cloud/core routes

Arqu:
I've cobbled together some wannabe fixes (my react foo is barely enough to make it work, best practices are out of the window). But this does set the stage for a few things to work better between core/cloud. Read up on the notes in the related issue for a better overview https://github.com/qri-io/qri/issues/1800 (especially the last section). 

~~Depends on: https://github.com/qri-io/cloud/pull/519 & https://github.com/qri-io/qri/pull/1816~~

### DIFF
--- a/src/chrome/PageControl.tsx
+++ b/src/chrome/PageControl.tsx
@@ -24,7 +24,7 @@ interface PageControlProps {
 
 // pass in pageInfo prop to show pagination info and controls
 const PageControl: React.FC<PageControlProps> = ({
-  pageInfo,
+  pageInfo = {},
   queryParams,
   showRange = true,
   onPageChange
--- a/src/features/dataset/state/datasetActions.ts
+++ b/src/features/dataset/state/datasetActions.ts
@@ -36,7 +36,7 @@ function fetchDataset (ref: QriRef): ApiAction {
     type: 'dataset',
     ref,
     [CALL_API]: {
-      endpoint: 'dataset_summary',
+      endpoint: 'ds/get',
       method: 'GET',
       segments: ref,
       map: mapDataset
@@ -55,7 +55,7 @@ function fetchBody (ref: QriRef, page: number, pageSize: number): ApiAction {
     type: 'body',
     ref,
     [CALL_API]: {
-      endpoint: 'get',
+      endpoint: 'ds/get',
       method: 'GET',
       pageInfo: {
         page,
--- a/src/features/dataset/state/datasetActions.ts
+++ b/src/features/dataset/state/datasetActions.ts
@@ -28,7 +28,7 @@ export function downloadLinkFromQriRef(ref: QriRef, body: boolean = false): stri
   if (body) {
     return `${API_BASE_URL}/ds/get/${ref.username}/${ref.name}${pathSegment}/body.csv`
   }
-  return `${API_BASE_URL}/download/${ref.username}/${ref.name}${pathSegment}`
+  return `${API_BASE_URL}/ds/get/${ref.username}/${ref.name}${pathSegment}?format=zip`
 }
 
 function fetchDataset (ref: QriRef): ApiAction {
--- a/src/features/dsPreview/state/dsPreviewActions.ts
+++ b/src/features/dsPreview/state/dsPreviewActions.ts
@@ -3,9 +3,14 @@ import { ApiAction, ApiActionThunk, CALL_API } from "../../../store/api";
 
 export function loadDsPreview(ref: QriRef): ApiActionThunk {
   return async (dispatch, getState) => {
-    return dispatch(fetchDsPreview(ref))
+    await Promise.all([dispatch(fetchDsPreview(ref))])
+    dispatch(fetchDsPreviewBody(ref))
+    dispatch(fetchDsPreviewReadme(ref))
+    return dispatch(fetchPreviewDone(ref))
   }
 }
+
+export const bodyPageSizeDefault = 50
 
 function fetchDsPreview (ref: QriRef): ApiAction {
   return {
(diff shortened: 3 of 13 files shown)


src/chrome/PageControl.tsx
```diff
@@ -29,6 +29,7 @@ const PageControl: React.FC<PageControlProps> = ({
   showRange = true,
   onPageChange
 }) => {
+  pageInfo = pageInfo || {}
```
Arqu: We don't have proper paging responses on core so this fails elsewhere with `pageInfo` being `undefined`. The downside to this fix is you see a bunch of "page 0 out of 0 with 0 results" kind of sections on the UI.

src/features/dataset/state/datasetActions.ts
```diff
@@ -55,7 +55,7 @@ function fetchBody (ref: QriRef, page: number, pageSize: number): ApiAction {
     type: 'body',
     ref,
     [CALL_API]: {
-      endpoint: 'get',
+      endpoint: 'ds/get',
```
Arqu, line 58: Think this was a merge issue.

src/features/dataset/state/datasetActions.ts
```diff
@@ -36,7 +36,7 @@ function fetchDataset (ref: QriRef): ApiAction {
     type: 'dataset',
     ref,
     [CALL_API]: {
-      endpoint: 'dataset_summary',
+      endpoint: 'ds/get',
```
Arqu, line 39: `ds/get` exists & works the same on core and cloud and should be used instead.

src/features/dsPreview/state/dsPreviewActions.ts
```diff
@@ -3,10 +3,15 @@ import { ApiAction, ApiActionThunk, CALL_API } from "../../../store/api";
 
 export function loadDsPreview(ref: QriRef): ApiActionThunk {
   return async (dispatch, getState) => {
-    return dispatch(fetchDsPreview(ref))
+    await Promise.all([dispatch(fetchDsPreview(ref))])
```
Arqu: This smells, but it gets the job done. We need to wait for the preview to load body and readme data (as it's not an all inclusive call) and they have to be added on to the preview.

src/features/dsPreview/state/dsPreviewActions.ts
```diff
@@ -3,10 +3,15 @@ import { ApiAction, ApiActionThunk, CALL_API } from "../../../store/api";
...
+    dispatch(fetchDsPreviewBody(ref))
+    dispatch(fetchDsPreviewReadme(ref))
+    return dispatch(fetchPreviewDone(ref))
   }
 }
 
+export const bodyPageSizeDefault = 50
```
Arqu, line 28: copy from `datasetActions`, we should think about just importing those here and reusing them.

src/features/dsPreview/state/dsPreviewState.ts
```diff
@@ -21,8 +21,28 @@ export const dsPreviewReducer = createReducer(initialState, {
     state.preview = undefined
     state.loading = true
   },
+  'API_DATASETPREVIEWFETCHDONE_REQUEST': (state) => {
```
Arqu: HELP

This is ugly but again makes things work. JS is weird.

src/features/search/Search.tsx
```diff
@@ -40,7 +40,7 @@ const Search: React.FC<{}> = () => {
   // if the query string ever changes, fetch new data
   useEffect(() => {
     dispatch(loadSearchResults(searchParams))
-  }, [q, page, sort, dispatch, searchParams])
+  }, [q, page, sort, dispatch])
```
Arqu: `searchParams` start a feedback loop and goes infinitely spamming search requests.

src/features/search/state/searchActions.ts
```diff
@@ -24,8 +24,11 @@ function fetchSearchResults (searchParams: SearchParams): ApiAction {
   return {
     type: 'search',
     [CALL_API]: {
-      endpoint: 'search',
-      method: 'GET',
+      endpoint: 'registry/search',
+      method: 'POST',
+      body: {
+        q: searchParams.q,
```
Arqu, line 33: We add this to the body, yet retain the url query param too. The reason being, core operates on our RPC dialect and requires a `POST` and `JS/TS` requires a non nil body so I just put something there, empty object didn't pass for me. 

On the other hand internally core uses both `GET` and `POST` against cloud to get the search results, where cloud treats pretty much all requests as GET. Should clean up the core search code and remove the GET version and transition it all to POST only I guess.

src/features/search/state/searchState.ts
```diff
@@ -34,7 +34,7 @@ export const searchReducer = createReducer(initialState, {
   'API_SEARCH_SUCCESS': (state: SearchState, action) => {
     state.results = action.payload.data
     state.loading = false
-    state.pageInfo = action.payload.pagination
+    state.pageInfo = action.payload.pagination || {}
```
Arqu, line 37: Another `undefined` pagination guard.

src/qri/search.ts
```diff
@@ -52,37 +52,38 @@ export interface SearchResult extends Dataset {
 }
 
 export function NewSearchResult(d: Record<string,any>): SearchResult {
-  const ds = NewDataset(d)
+  const dv = d.value
+  const ds = NewDataset(dv)
   let stats
-  let followStats
-  let issueStats
+  let followStats = {}
```
Arqu, line 58: core does not supply follow or issue stats and fails calls bellow this.

src/utils/commitishFromPath.ts
```diff
@@ -1,3 +1,6 @@
 export default function commitishFromPath (path: string): string {
-  return path.split('/')[2].substr(0,8)
+	if (path) {
```
Arqu, line 2: Unsure how bad this really is, but had a lot of catastrophic failures on this because we returned bad data. 
While this is not a good fix, it does bring to light that this is one of many places where we have the frontend explode on the slightest deviation from expectations.

Arqu (commented):
Round of self review to highlight sections and issues.

src/chrome/PageControl.tsx
b5: interesting. I think we should do two things:
1. get an issue up on core to checklist out all api endpoints that need page info but don't have it
2. this line is still good defensive coding, but can be refactored as a default value in the arguments list:

```tsx
const PageControl: React.FC<PageControlProps> = ({
  pageInfo = {},
}) => {}
```

src/features/dataset/state/datasetActions.ts
b5: ageree. I think we should go so far as to remove `dataset_summary` and mandate that `ds/get` without a selector returns a preview

src/features/dsPreview/state/dsPreviewActions.ts
b5: This is good for now. We can work on this on the backend

src/features/dsPreview/state/dsPreviewState.ts
b5: hrm, yeah let's deal with the rest & then I'll take over this PR & clean this up

src/features/search/state/searchActions.ts
b5: sounds great. Mind adding a note in a comment here?

src/qri/search.ts
b5: interesting. ok we may need to adjust frontend code to not require these be defined

src/utils/commitishFromPath.ts
b5: interesting. this'll work for now. We should clean this up in the backend

PR #208, opened 2021-06-24
dry run hotkey, move deploy button

chriswhong:
- Adds cmd+enter/ctrl+enter hotkey to trigger dry run
- Splits the deploybutton, leaving a description in the workflow outline, and placing the button at the bottom of the workflow editor.

PR #209, opened 2021-06-24
workflow modals

chriswhong:
Builds on top of work in #208, which should be merged first.

- Adds `disabled` state to Button, which is then used in `WorkflowEditor` to disable the deploy button until a dry run has succeeded.

- Adds two modals:

  - Workflow splash modal, appears when user first lands on the workflow editor, includes a carousel of text-based tips
  - Deploy modal, appears when the user clicks deploy in the workflow editor, prompts the user to change the dataset name (for now, assumes the name always needs to be changed). Need to determine the correct sequence of events when user finally hits deploy (what do we do with the user-provided new name?)

- Adds a "+" icon to the Triggers step in workflows, removes default trigger from new workflows (big TODO to add UI for making triggers)
- Adds the keyboard shortcut to the dry run button tooltip
- Replaces `CloseButton` with `IconButton`, useful for icons that act like buttons but have no background/border
- Removes row of `DatasetInfoItem` from `DatasetHeader` when editing a new workflow. 
- Changes 'New Version Preview'  to 'output' in the workflow editor
-


b5:
#208 is in, mind rebasing?

chriswhong:
Rebased

src/chrome/Button.tsx
```diff
@@ -23,13 +24,17 @@ const Button: React.FC<ButtonProps> = ({
...
+  disabled = false,
   children
 }) => (
   <button
     type={submit ? 'submit' : 'button'}
     className={classNames(
       'inline-flex items-center justify-center rounded-md shadow-sm bg-transparent font-medium focus:outline-none focus:ring focus:ring-offset ring-offset-transparent mt-0 transition-all duration-100',
       className,
+      {
+        'cursor-default bg-opacity-40 hover:bg-opacity-40': disabled
```
b5, line 36: should this be `cursor-cancel`?

src/features/app/App.tsx
```diff
@@ -7,9 +7,13 @@ import Routes from '../../routes'
 import Modal from './modal/Modal'
 import { wsConnect } from '../websocket/state/websocketActions';
 
-import './App.css';
 import SnackBar from '../snackBar/SnackBar';
 
+import './App.css';
+// import styles for react-responsive-carousel, used in workflow splash modal
+import 'react-responsive-carousel/lib/styles/carousel.min.css'
```
b5: move this import into `src/features/modal/SplashModal.tsx`. CSS imports should be colocated with the component that utilizes them

src/chrome/Button.tsx
```diff
@@ -45,7 +50,8 @@ const Button: React.FC<ButtonProps> = ({
         'text-white bg-red-600 hover:bg-red-700 focus:ring-red-500': (type === 'danger'),
       }
     )}
-    onClick={onClick}
+    onClick={disabled ? () => {} : onClick}
```
b5: do we need to pass a noop function here? seems like `onClick={!disabled && onClick}` would make more sense

src/chrome/IconButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+import Icon, { IconSize } from './Icon'
+
+interface IconButtonProps {
+  className?: string
+  icon: string
+  size?: IconSize
+  onClick: () => void
+}
+
+const IconButton: React.FunctionComponent<IconButtonProps> = ({
```
b5: React.FC

src/features/collection/Collection.tsx
```diff
@@ -90,7 +92,13 @@ const Collection: React.FC<{}> = () => {
 
             <div className='w-1/2 flex items-center justify-end'>
               <SearchBox onChange={handleSearchChange} placeholder='Filter' dark />
-              <Link to='/ds/new'>
+              <Link to={{
+                pathname: `/ds/${username}/dataset_${Math.floor(Math.random() * 1000)}/workflow/new`,
```
b5, line 96: we can tackle this in a follow up, but we should drop the random dataset name generation

src/features/workflow/RunBar.tsx
```diff
@@ -32,6 +32,8 @@ const RunBar: React.FC<RunBarProps> = ({
 
   const handleCancel = () => { alert('cannot cancel runs yet') }
 
+  const isMac = navigator.platform.indexOf('Mac') > -1
```
b5: move this to a new file `utils/platform.ts` & wrap it in an exported function:

```ts
export function platform(): 'mac' | 'pc' {
  if (navigator.playform.indexOf('Mac') > -1) {
    return 'mac'
  }
  return 'pc'
}
```

replace this line with: `const isMac = (platform() === 'mac')`

src/features/workflow/WorkflowPage.tsx
```diff
@@ -48,6 +48,12 @@ const WorkflowPage: React.FC<WorkflowPageProps> = ({
     qriRef.username = user.username
   }
 
+  // determine if the workflow is new by reading /new at the end of the pathname
+  const segments = useLocation().pathname.split('/')
+  const isNewWorkflow = segments[segments.length - 1] === 'new'
+
+
+
```
b5: no need for these spaces

src/features/workflow/modal/SplashModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+      </div>
+
+      <Carousel
+        showStatus={false}
+        showArrows={false}
+        showThumbs={false}
+        interval={3000}
+        showIndicators
+        infiniteLoop
+        autoPlay
```
b5: should this autoPlay? Most examples I've seen of this UX in the wild require me to scroll

b5 (approved):
Looks great! a few nits then let's merge

src/chrome/Button.tsx
chriswhong: Don't think so, `cursor-cancel` isn't an option in the tailwind docs.  https://tailwindcss.com/docs/cursor

src/chrome/Button.tsx
chriswhong: `onClick={!disabled && onClick}` causes typescript to yell.  I changed it to `onClick={onClick}` and added the noop as the default value for `onClick`

Issue #212, opened 2021-06-25
UI Punchlist Items


chriswhong:
- [x] When user is on a New Workflow, other dataset nav items should be grayed out (Preview, history, run log)
- [x] Too much padding on left and right of output body preview
- [x] Mini Header should instantly disappear when user scrolls back to the top, not animate up
- [x] Need some better placeholder for empty output dataset.  Right now it just says “such empty”
- [x] Expanding Dataset Nav leaves code editor cells at their previous width (need to call the method that resizes them to their container)
- [x] Icons do a slight vertical jump when expanding/closing dataset nav
- [x] Clicking output dataset in outline shows navigation warning (because it thinks it’s a link in addition to being a scrolltrigger)
- [x] Let’s show a tooltip over DeploymentStatus in workflow outline
- [x] Empty triggers and on completion in outline should show something, perhaps ‘--’
- [x] Collection view should prioritize name column when narrowed
- [x] Let’s say “Result” instead of “Output”
- [x] Footer Icon Links don’t work

PR #222, opened 2021-06-28
Cw/UI punchlist

chriswhong:
Fixes ALL THE THINGS in #212:

- [x] When user is on a New Workflow, other dataset nav items should be grayed out (Preview, history, run log)
- [x] Too much padding on left and right of output body preview
- [x] Mini Header should instantly disappear when user scrolls back to the top, not animate up
- [x] Need some better placeholder for empty output dataset.  Right now it just says “such empty”
- [x] Expanding Dataset Nav leaves code editor cells at their previous width (need to call the method that resizes them to their container)
- [x] Icons do a slight vertical jump when expanding/closing dataset nav
- [x] Clicking output dataset in outline shows navigation warning (because it thinks it’s a link in addition to being a scrolltrigger)
- [x] Let’s show a tooltip over DeploymentStatus in workflow outline
- [x] Empty triggers and on completion in outline should show something, perhaps ‘--’
- [x] Collection view should prioritize name column when narrowed
- [x] Let’s say “Result” instead of “Output”
- [x] Footer Icon Links don’t work

### DIFF
--- a/src/features/collection/CollectionTable.tsx
+++ b/src/features/collection/CollectionTable.tsx
@@ -1,4 +1,4 @@
- import React, { forwardRef } from 'react'
+import React, { forwardRef } from 'react'
 import { useDispatch } from 'react-redux';
 import numeral from 'numeral'
 import ReactDataTable from 'react-data-table-component'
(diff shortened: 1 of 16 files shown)


src/features/collection/CollectionTable.tsx
```diff
@@ -188,19 +188,20 @@ const CollectionTable: React.FC<CollectionTableProps> = ({
...
-      },
-      selector: 'triggers',
-      omit: simplified,
-      width: '160px',
-      cell: (row: VersionInfo) => (row.id
-          ? <div className='tracking-wider font-medium text-qrinavy'>Schedule, Run When, Webhook</div>
-          : '—'
-      )
-    },
+    // let's hide the Triggers column for now, to provide more room for the name column on narrow screens
```
b5, line 191: 👍

b5 (approved):
LGTM

PR #224, opened 2021-06-29
Cw/bugfixes

### LINKED ISSUE #67: Workflow Editor has status icons from previous dataset
After a dry run produced checkmarks in the workflow editor, go to collection view, open another dataset, and the checkmarks are still there.  We need to make sure the state is cleared and reflects the current dataset.

### LINKED ISSUE #178: Implement full screen body
Make the full screen button actually work!

### LINKED ISSUE #211: Unnecessary API call to dataset_summary when opening new workflow
New Workflow still calls `dataset_summary`, we should not make this call if there is /new in the URL (or however else)

### LINKED ISSUE #223: User can Remove a dataset from the collection view
- Add 'Remove" as an option
  - Show an "are you sure" modal

chriswhong:
- User can remove a dataset from Collection View Row Menu (Closes #223)
- Don't fetch data on new workflows (Closes #211)
- Prevent Body fetch on Workflow Dry Run Preview (also related to #211)
- Fix full screen body layout when triggered from Dry Run Preview (Closes #178)
- Reset workflow state on `API_WORKFLOW_REQUEST` (Closes #67)

### DIFF
--- a/src/features/dataset/state/datasetActions.ts
+++ b/src/features/dataset/state/datasetActions.ts
@@ -1,5 +1,5 @@
 import Dataset, { NewDataset } from "../../../qri/dataset"
-import { QriRef } from "../../../qri/ref"
+import { QriRef, refStringFromQriRef } from "../../../qri/ref"
 import { ApiAction, ApiActionThunk, CALL_API } from "../../../store/api"
 import { RENAME_NEW_DATASET } from "./datasetState"
 import { API_BASE_URL } from '../../../store/api'
@@ -80,7 +80,7 @@ export function removeDataset (ref: QriRef): ApiActionThunk {
         endpoint: 'ds/remove',
         method: 'POST',
         body: {
-          ref: `${ref.username}/${ref.name}`
+          ref: refStringFromQriRef(ref)
         }
       }
     }
--- a/src/features/dataset/state/datasetActions.ts
+++ b/src/features/dataset/state/datasetActions.ts
@@ -77,12 +77,10 @@ export function removeDataset (ref: QriRef): ApiActionThunk {
     const action = {
       type: 'remove',
       [CALL_API]: {
-        endpoint: 'remove',
-        method: 'DELETE',
-        segments: {
-          peername: ref.username,
-          name: ref.name,
-          path: ref.path
+        endpoint: 'ds/remove',
+        method: 'POST',
+        body: {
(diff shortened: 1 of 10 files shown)


src/features/dataset/state/datasetActions.ts
```diff
@@ -77,12 +77,10 @@ export function removeDataset (ref: QriRef): ApiActionThunk {
...
-        endpoint: 'remove',
-        method: 'DELETE',
-        segments: {
-          peername: ref.username,
-          name: ref.name,
-          path: ref.path
+        endpoint: 'ds/remove',
+        method: 'POST',
+        body: {
+          ref: `${ref.username}/${ref.name}`
```
b5: import & use `refStringFromQriRef`

src/features/dsComponents/DatasetComponent.tsx
```diff
@@ -16,11 +16,14 @@ import IconLink from '../../chrome/IconLink'
 export interface DatasetComponentProps {
   dataset: Dataset
   componentName: ComponentName
+  // preview will cause the body component to render only what is in dataset and not fetch more data
+  preview?: boolean
```
b5, line 20: not a massive fan of this "preview" prop name, but given that it's already landed in the `Body` component, let's stick with it. I'd prefer `noFetch`, which would make the behavior it's affecting more accurate

src/features/workflow/WorkflowEditor.tsx
```diff
@@ -53,6 +62,23 @@ const WorkflowEditor: React.FC<WorkflowEditorProps> = ({ runMode, run, workflow
...
 
+  // appends username, name, and meta.title to a dry run's preview, useful for
+  // display of downstream components that expect these as part of a Dataset (e.g. fullscreen body)
+  const appendRefAndMeta = (dsPreview: Dataset) => {
+    if (dsPreview) {
+      return {
+        peername: qriRef.username,
+        name: qriRef.name,
+        meta: {
+          title: 'Dry Run Data Preview'
```
b5: what about "Workflow Result"? What happens if the user writes a transform script that does this?

```py
def transform(ds, ctx):
  ds.set_meta("title", "some other title")
```

I'd say change the `meta` assignment to:

```tsx
return {
 // ...
  meta: dspreview.meta || { title: "Workflow Result" }
}
```

b5 (approved):

PR #231, opened 2021-07-09
Workflow Deploy UI/UX

### LINKED ISSUE #74: Deployment UX
Deploy button should not be clickable until a dry run succeeds, should disable again when code is in a dirty state.  Also, dirty/deployed state should determine whether a warning is shown when user navigates away.

chriswhong:
- Adds a "Run on Deploy" checkbox to `DeployModal` which sets the `run` property in the call to `auto/deploy`
- Renames `onComplete` to `hook` in several places
- Fixes a bug where the active item in `Select` was not styled properly
- Improves new workflow UX
  - New workflow route becomes `/workflow/new`, other dataset nav options are grayed out
  - Improved logic for calculating workflow dirty state
  - Unsaved changes modal fires at the correct times (on isDirty) and behaves as expected
  - Improves and wires up deployModal, on deploy user sees loading spinners that accurately reflect the deploy status.
  - On deploy end, the route is changed to `/ds/username/name/workflow`, the deployed workflow is a bona fide dataset!
  - Resets run events on edit (resets a successful or failed dry run)
  - User cannot reach deploy modal unless they have dirty state and a successful dry run
  - Handles both new workflows and updates of existing workflows!

Closes: #74


b5 (approved):
I _really_ need these PRs to get smaller to be able to review them properly. For the sake of keeping moving, let's merge this. But let's chat about strategies for cutting these down. PRs that are easier to review will sit around for less time.

PR #232, opened 2021-07-09
fix: use ref to store workflow for dry run keyboard shortcut execution

### LINKED ISSUE #229: Keyboard Shortcut for dry run shows success when script should fail
When working on a workflow, I am seeing a situation where Cmd-enter (mac) shows the running state and checkmarks when the script should be failing.  Clicking Dry Run manually results in the expected failure and print outputs, but using the shortcut seems to not actually run the script, clears the outputs (but the preview stays the same) and shows checkmarks for each step.

chriswhong:
Uses a `ref` to store the `workflow` prop in `WorkflowEdtor`.  Fixes a closure issue where the `workflow` used when starting dry run with a keyboard shortcut was the `initialState` of `workflow` from the redux store.  

Closes #229

### DIFF
--- a/src/features/workflow/WorkflowEditor.tsx
+++ b/src/features/workflow/WorkflowEditor.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useState, useRef, useEffect } from 'react'
 import { useDispatch } from 'react-redux'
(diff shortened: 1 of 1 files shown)


b5 (approved):
LGTM

PR #234, opened 2021-07-29
feat: improve rename UI in datasetHeader

chriswhong:
Improves the rename UI in `DatasetHeader`, adding:
- press escape to stop editing
- click outside the input to stop editing
- validate user input for a datasetname (don't allow user to enter an invalid name)
- auto-select text when editing starts so the user can start typing immediately
- ensure route update and re-flow of data after name change


chriswhong:
superseded by #236

PR #236, opened 2021-08-03
feat: improve rename UI in datasetHeader

chriswhong:
Fixes/improves the rename UI (user can rename a dataset they own by clicking the dataset name in any of the dataset views)

Changes the header when the user is creating a new dataset from workflow

### DIFF
--- a/src/features/dataset/DatasetHeader.tsx
+++ b/src/features/dataset/DatasetHeader.tsx
@@ -44,7 +44,6 @@ const DatasetHeader: React.FC<DatasetHeaderProps> = ({
       .then(({ type }) => {
         if (type === 'API_RENAME_SUCCESS') {
           const newPath = history.location.pathname.replace(dataset.name, value)
-          console.log('routing from ', history.location.pathname, ' to ', newPath)
           history.replace(newPath)
         }
       })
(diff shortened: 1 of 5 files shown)


src/features/workflow/WorkflowPage.tsx
```diff
@@ -58,7 +58,7 @@ const WorkflowPage: React.FC<WorkflowPageProps> = ({ qriRef }) => {
     dispatch(loadDataset(ref))
     dispatch(loadWorkflowByDatasetRef(qriRef))
 
-  }, [])
+  }, [ qriRef ])
```
b5, line 61: looks good, but a note I want to make sure you're aware of when passing a `qriRef` as a dep to hooks like `useEffect`: as written any change to the `qriRef` _object_ will cause a re-render, so if anything up the stack is, say, constructing the qri ref from URL params, that'll trigger a re-render. One way to avoid that is to spread the properties into the hook memoization array:

```tsx
const { username, name, path } = qriRef
useEffect(() => {
  const ref = newQriRef({ username, name, path})
  // ...
}, [username, name, path])
```

The way you have it written is totally fine, but if you find you're getting weird re-rendering, it's possible that the cause is reconstruction of a new qriRef object with the same properties. If that happens, switch to this form

src/features/dataset/state/datasetActions.ts
```diff
@@ -96,7 +91,7 @@ export interface RenameDatasetAction {
   next: QriRef
 }
 
-export function renameDataset(current: QriRef, next: QriRef): ApiActionThunk | Promise<RenameDatasetAction> {
+export function renameDataset(current: QriRef, next: QriRef): ApiActionThunk {
```
b5, line 94: lovely

src/features/dataset/DatasetHeader.tsx
```diff
@@ -38,39 +40,51 @@ const DatasetHeader: React.FC<DatasetHeaderProps> = ({
...
-    // TODO(b5): we should be chaining this route replacement after successful
-    // dispatch with a "then" off the renameDataset action
-    const newPath = history.location.pathname.replace(dataset.name, value)
-    console.log('routing from ', history.location.pathname, ' to ', newPath)
-    history.replace(newPath)
+    renameDataset(qriRef, { username: dataset.peername, name: value })(dispatch)
+      .then(({ type }) => {
+        if (type === 'API_RENAME_SUCCESS') {
+          const newPath = history.location.pathname.replace(dataset.name, value)
+          console.log('routing from ', history.location.pathname, ' to ', newPath)
```
b5: stray console log

b5 (approved):
one stray console.log, then lgtm!

src/features/workflow/WorkflowPage.tsx
chriswhong: 👍 I think this is actually cleaned up in some of my follow-on work, but I'll keep an eye out for it.

PR #238, opened 2021-08-05
feat: add toasts for runs

chriswhong:
Adds toasts that appear when a run starts, updates on success, disappears after a few seconds

### DIFF
--- a/src/features/toast/ToastRenderer.tsx
+++ b/src/features/toast/ToastRenderer.tsx
@@ -1,0 +1,80 @@
+import React, { useRef } from 'react'
+import { useSelector } from 'react-redux'
+import { ToastContainer, toast } from 'react-toastify'
+
+import { Run, RunStatus } from '../../qri/run'
+import Toast from './Toast'
+import { selectRunsFromEventLog } from '../workflow/state/workflowState'
+
+
+import 'react-toastify/dist/ReactToastify.css'
+
+const ToastRenderer: React.FC<{}> = () => {
+
+  const runs = useSelector(selectRunsFromEventLog)
+
+  // use a ref to store the workflow status so we can
+  // determine when it changes and update an already visible toast
+  const runsRef = useRef({} as Record<string, RunStatus>)
+
+  // artifically delay the update to 'success', ensures that
+  // user will see 'running' for at least 3 seconds even if the
+  // workflow runs quickly
+  const SUCCESS_START_DELAY = 3000
+
+  // after 'success', close the toast 4 seconds later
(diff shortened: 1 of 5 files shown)


src/features/workflow/state/workflowState.ts
```diff
@@ -28,6 +29,14 @@ export const selectLatestRun = (state: RootState): Run | undefined => {
   return undefined
 }
 
+export const selectRunsFromEventLog = (state: RootState): any => {
```
ramfox: does this need to return `any`, should the type be `Run[]`?

src/features/toast/ToastRenderer.tsx
```diff
@@ -0,0 +1,80 @@
...
+              render: (
+                <Toast message='Success!' type='succeeded' id={id}  />
+              ),
+            })
+            setTimeout(() => {
+              delete runsRef.current[id]
+              toast.dismiss(id)
+            }, SUCCESS_START_DELAY)
+          }, SUCCESS_CLOSE_DELAY)
+        }
```
ramfox: rather than having a `setTimeout` to dismiss the toast, shouldn't we instead pass in a `autoClose` value of `SUCCESS_CLOSE_DELAY` and an an `onClose` callback that deletes the id from the `runsRef` object?

```
if (runsRef.current[id] === 'running') {
    runsRef.current[id] = status
    setTimeout(() => {
        toast.update(id, {
            render: <Toast message='Success!' type='succeeded' id={id}  />,
            autoClose: SUCCESS_CLOSE_DELAY,
            onClose: () => { delete runsRef.current[id] }
        })}, SUCCESS_START_DELAY)
}
```

The `setTimeout` to delay the start I can't think of a less complicated solution for the moment.


Regardless,`SUCCESS_START_DELAY` is supposed to be the time before we show the "Success", and `SUCCESS_CLOSE_DELAY` is for the time we see the "Success" toast before it closes, correct? If so, these are backwards in the code.

ramfox (approved):
LGTM

My one nit is around the `setTimeout`s, which I find confusing to stack and think we should switch to using a callback & the `autoClose` timer instead.

src/features/workflow/state/workflowState.ts
chriswhong: Good catch!

src/features/toast/ToastRenderer.tsx
chriswhong: Tried this method out, it works great and is much easier to read.

PR #239, opened 2021-08-05
feat: style run log, refactor dataset page layout

chriswhong:
- style/layout improvements to the dataset run log, making it match the mockups and appear more like the collection list

- refactor dataset view components, consolidating duplicate code and introducing layout components, fixed (history, run log) and scrolling (preview, workflow)

- style tweaks for collection table, commit list, and commit header

### DIFF
--- a/src/chrome/Icon.tsx
+++ b/src/chrome/Icon.tsx
@@ -1,49 +1,5 @@
 import React from 'react'
 
-import { FontAwesomeIcon, FontAwesomeIconProps } from '@fortawesome/react-fontawesome'
-import {
-  faArrowLeft,
-  faArrowRight,
-  faBars,
-  faBolt,
-  faCheckCircle,
-  faCircle,
-  faCloudUploadAlt,
-  faCode,
-  faProjectDiagram,
-  faEllipsisH,
-  faEnvelope,
-  faFont,
-  faHashtag,
-  faHome,
-  faHdd,
-  faList,
-  faCheck,
-  faPen,
-  faMinusCircle,
-  faPlay,
-  faPlus,
-  faPlusCircle,
-  faPauseCircle,
-  faTimes,
-  faToggleOn,
-  faExclamationCircle,
-  faExclamationTriangle,
-  faQuestion,
-  faQuestionCircle,
-  faSearch,
-  faShip,
-  faSortDown,
-  faSpinner,
-  faTable,
-
-  IconDefinition,
-} from '@fortawesome/free-solid-svg-icons'
-
-import {
-  faFile
-} from '@fortawesome/free-regular-svg-icons'
-
 import ActivityFeed from './icon/ActivityFeed'
 import AutomationFilled from './icon/AutomationFilled'
 import Body from './icon/Body'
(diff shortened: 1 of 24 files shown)


src/features/dsPreview/state/dsPreviewActions.ts
```diff
@@ -7,17 +7,17 @@ export function loadDsPreview(ref: QriRef): ApiActionThunk {
 
     try {
       await Promise.all([
-        dispatch(fetchDsPreview(ref)),
+        // dispatch(fetchDsPreview(ref)),
```
ramfox, line 12: stray comment

src/chrome/Icon.tsx
```diff
@@ -1,49 +1,5 @@
...
-  faSpinner,
-  faTable,
-
-  IconDefinition,
-} from '@fortawesome/free-solid-svg-icons'
-
-import {
-  faFile
-} from '@fortawesome/free-regular-svg-icons'
-
```
ramfox, line 46: guessing this is because we now have custom icons slash have a narrower list of icons we are expected to show?

src/features/dsPreview/state/dsPreviewActions.ts
ramfox: hmmmm, is the idea here that because we now load previews to populate the dataset headers, not to fetch them again in the `loadDsPreview`?

If so, I would vote for adding back in `dispatch(fetchDsPreview(ref))`, or at least adding a TODO or some commentary. I would be confused in the future when using `loadDsPreview` that it doesn't actually load a bunch of the dataset.

src/features/dataset/DatasetScrollLayout.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+import { newQriRef } from '../../qri/ref'
+import { Dataset } from '../../qri/dataset'
+import { selectDsPreview } from '../dsPreview/state/dsPreviewState'
+import { selectSessionUser } from '../session/state/sessionState'
+import { selectSessionUserCanEditDataset } from './state/datasetState'
+import DatasetHeader from './DatasetHeader'
+import DatasetMiniHeader from '../dataset/DatasetMiniHeader'
+
+interface DatasetFixedLayoutProps {
```
ramfox: should this be `DatasetScrollLayoutProps`?

src/features/dataset/DatasetScrollLayout.tsx
```diff
@@ -0,0 +1,70 @@
...
+  dataset?: Dataset
+  headerChildren?: JSX.Element
+  contentClassName?: string
+}
+
+const DatasetFixedLayout: React.FC<DatasetFixedLayoutProps> = ({
```
ramfox: ? `DatasetScrollLayout` & `DatasetScrollLayoutProps`

src/features/dataset/DatasetScrollLayout.tsx
```diff
@@ -0,0 +1,70 @@
...
+          {children}
+        </div>
+      </div>
+    </>
+  )
+}
+
+
+
+export default DatasetFixedLayout
```
ramfox: ?
```
export default DatasetScrollLayout
```

ramfox (approved):
LGTM, the dataset layouts are much clearer!

just a few renaming/commenting requests

src/features/dataset/DatasetScrollLayout.tsx
chriswhong: Good catch

src/chrome/Icon.tsx
chriswhong: Yes, all icons should be our custom icons now.

PR #241, opened 2021-08-05
No unused UI

chriswhong:
Merge #239 first

- Comments out unused UI elements (Dashboard link in header, search box above history list, etc) which were mocked up but did not function
- Improves contextual buttons on dataset routes:
  - Preview gets a Download Button
  - History get no button
  - Workflow gets a Dry Run Button
  - Run Log gets a Run Now Button
- Make version info box on previews match version info header in history view 
- Removes fontawesome icons which should no longer be in use in the app


ramfox (approved):
LGTM

PR #242, opened 2021-08-09
Data types

chriswhong:
Merge #241 first

- Adds custom icons for data type glyphs, and cleans up the layout/colors for the two places data type glyphs are used (Body table column headers and structure)
- Restores the missing `more` (three dots) icon used in collection view
- Improve `RelativeTimstamp` custom label abbreviations to handle day/days
- Improves the layout of Dataset Preview's body section (was a long table, now screen height with scrolling table rows)


ramfox (approved):
LGTM

yay icons in body headers!

I don't hate it, because it helps the body be the focus of the whole screen, but just a note that we do now have a double scroll situation on preview

PR #243, opened 2021-08-10
feat: auto refresh access token

chriswhong:
Modifies API middleware to check the user's `accessToken` on all requests.  If it expires in the next 5 minutes, use the `refreshToken` to request a new `accessToken` and update state/localstorage with the new value.

Includes a TODO for delaying subsequent requests while a refresh is happening (pages that fire multiple requests when a token is near expiration will fire multiple calls to refresh to the token)

### DIFF
--- a/src/store/api.ts
+++ b/src/store/api.ts
@@ -1,10 +1,11 @@
 import { Middleware, Dispatch, AnyAction } from 'redux'
 import { ThunkDispatch } from 'redux-thunk'
+import jwtDecode from 'jwt-decode'
 
 import { RootState } from './store'
 import mapError from './mapError'
 import { QriRef } from '../qri/ref'
-import { selectSessionToken } from '../features/session/state/sessionState'
+import { selectSessionTokens } from '../features/session/state/sessionState'
 
 // CALL_API is a global, unique constant for passing actions to API middleware
 export const CALL_API = Symbol('CALL_API')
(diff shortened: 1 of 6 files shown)


src/store/api.ts
```diff
@@ -300,3 +310,71 @@ export const apiMiddleware: Middleware = ({ getState }) => (next: Dispatch<AnyAc
...
+
+        return Promise.resolve(res.data.access_token)
+      }
+    }
+  } catch (error) {
+    console.log(`error refreshing token - ${error.message}`)
+    if (token) {
+      return Promise.resolve(token)
+    }
+    // maybe invalidate session?
```
ramfox: Do we have logic for re-routing to the login page if the user is on or tries to navigate to a private route without having a valid token? If so, this is where we should reroute them, or at least, catch this error higher up in the chain & send them to the login page.

chriswhong: Nice catch, I will work this in

ramfox:
I think we are still missing something. Except for `workflow/new` & the login page, we should probably have the app or each route wrapped. If there is no proper session id then we should reroute to the login page. I thought I saw something like this in the code already in `routes.ts`, `PrivateRoute`?

I guess I also am not totally clear on what a user is allowed to do while not logged in. Can they view datasets? Or is it just edit a new workflow (until we get to dryrun/deploy). If they are only permitted to edit a workflow, than we need to add further restrictions if they attempt to navigate directly to a dataset page.

chriswhong:
Yes, we have `PrivateRoute` wrapping Collection (and two other top-level routes that don't exist ATM), but everything else in a dataset is public for now.  We should probably implement some kind of read-only UI for workflows, or gray out and redirect from the workflow editor if the user is not logged in, but we can deal with that later I think.

ramfox (approved):
LGTM

We know we have more work to do around tokens and the non `workflow/new` parts of the dataset routes, but that will be for a later date!

PR #244, opened 2021-08-11
Refactor workflow state

chriswhong:
- Cleans up interface for `Workflow` to match backend. 
- Introduces `workflowDataset` into state, which is populated with HEAD but also where edits to transform steps can be mutated when the user is editing.  This is sent along with `workflow` when deploying.

- changes the layout of commits in the Run Log to match the commit layout used in preview/history
- Elevates the api call to workflow to `DatasetWrapper` and sets `hasWorkflow` in dsPreview so we always know if the current dataset has a workflow

### DIFF
--- a/src/features/workflow/state/workflowState.ts
+++ b/src/features/workflow/state/workflowState.ts
@@ -77,7 +77,7 @@ const initialState: WorkflowState = {
   runMode: 'apply',
   workflow: {
     id: '',
-    initID: 'new_dataset',
+    initID: '',
     active: true,
 
     triggers: [],
@@ -114,7 +114,6 @@ export const workflowReducer = createReducer(initialState, {
   SET_WORKFLOW_REF: (state, action: SetWorkflowRefAction) => {
     state.dataset.username = action.qriRef.username
     state.dataset.name = action.qriRef.name
-    state.workflow.initID = `${action.qriRef.username}/${action.qriRef.name}`
   },
   // listen for dataset fetching actions, if the reference of the fetched dataset
   // matches the ref the workbench reducer is tuned to, load the transform script
--- a/src/features/workflow/state/workflowState.ts
+++ b/src/features/workflow/state/workflowState.ts
@@ -2,17 +2,26 @@ import { createReducer } from '@reduxjs/toolkit'
 import DeepEqual from 'deep-equal'
 import _ from 'lodash'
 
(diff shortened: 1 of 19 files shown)


src/qri/dataset.ts
```diff
@@ -23,15 +23,15 @@ export default Dataset
 
 export function qriRefFromDataset(dataset: Dataset): QriRef {
   return {
-    username: dataset.peername,
+    username: dataset.peername || dataset.username,
```
ramfox, line 26: ooOooo exciting!!!

src/features/workflow/state/workflowState.ts
```diff
@@ -63,20 +77,20 @@ const initialState: WorkflowState = {
   runMode: 'apply',
   workflow: {
     id: '',
-    datasetID: 'fake_id',
-    runCount: 0,
+    initID: 'new_dataset',
```
ramfox: does `initID` need 'new_dataset'? If it can be an empty string, it would probably be better in the long term, I can see if accidentally intro-ing a bug at some point.

src/features/workflow/state/workflowState.ts
```diff
@@ -85,34 +99,31 @@ export const workflowReducer = createReducer(initialState, {
...
-  SET_WORKFLOW: setWorkflow,
   WORKFLOW_CHANGE_TRIGGER: changeWorkflowTrigger,
   WORKFLOW_CHANGE_TRANSFORM_STEP: changeWorkflowTransformStep,
   RUN_EVENT_LOG: addRunEvent,
   SET_WORKFLOW_REF: (state, action: SetWorkflowRefAction) => {
-    state.qriRef = action.qriRef
-    state.workflow.datasetID = `${action.qriRef.username}/${action.qriRef.name}`
+    state.dataset.username = action.qriRef.username
+    state.dataset.name = action.qriRef.name
+    state.workflow.initID = `${action.qriRef.username}/${action.qriRef.name}`
```
ramfox: the initID is no longer the `username/name` combination (we used to rely on the ref when we didn't have initID's in place). Unless we are relying on this for some reason, it would be better to have this string be empty unless we have an actual initID.

ramfox (approved):
LGTM, only questions are surrounding the `workflow.InitID` as it is stored in the workflow state.

PR #245, opened 2021-08-11
Run now

chriswhong:
Merge #244 first

- Wires up `Run Now` button in Run Log
- calculates `isDirty` on trigger changes (previously was only updating on changes to steps
- introduce `latestDryRun` to workflow, allowing us to select the dry run in the workflow editor, and a regular run in Run Log
- Adds small number UI to Dataset Nav Items, useful for showing version count, etc
- Consolidates commits and logs api calls to use the same action type, so that the commit list can be updated at the same time as the activity list

### DIFF
--- a/src/features/activityFeed/state/activityFeedActions.ts
+++ b/src/features/activityFeed/state/activityFeedActions.ts
@@ -14,13 +14,14 @@ export function loadDatasetLogs(qriRef: QriRef): ApiActionThunk {
 
 function fetchDatasetLogs(qriRef: QriRef): ApiAction {
   return {
-    type: 'dataset_logs',
+    type: 'dataset_activity',
     qriRef,
     [CALL_API]: {
       endpoint: `ds/activity`,
       method: 'POST',
       body: {
-        ref: `${qriRef.username}/${qriRef.name}`
+        ref: `${qriRef.username}/${qriRef.name}`,
+        pageSize: 100
       },
       map: mapLogs
     }
--- a/src/features/workflow/Workflow.tsx
+++ b/src/features/workflow/Workflow.tsx
@@ -6,7 +6,7 @@ import { Location } from 'history'
 
 import WorkflowOutline from './WorkflowOutline'
 import {
-  selectLatestRun,
+  selectLatestDryRun,
   selectRunMode,
   selectWorkflow,
   selectWorkflowIsDirty,
@@ -34,7 +34,7 @@ const Workflow: React.FC<WorkflowProps> = ({ qriRef }) => {
   const location = useLocation<WorkflowLocationState>()
   const workflow = useSelector(selectWorkflow)
   const workflowDataset = useSelector(selectWorkflowDataset)
-  const latestRun = useSelector(selectLatestRun)
+  const latestRun = useSelector(selectLatestDryRun)
   const runMode = useSelector(selectRunMode)
   const isDirty = useSelector(selectWorkflowIsDirty)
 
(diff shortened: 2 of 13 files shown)


src/features/workflow/Workflow.tsx
```diff
@@ -34,7 +34,7 @@ const Workflow: React.FC<WorkflowProps> = ({ qriRef }) => {
   const location = useLocation<WorkflowLocationState>()
   const workflow = useSelector(selectWorkflow)
   const workflowDataset = useSelector(selectWorkflowDataset)
-  const latestRun = useSelector(selectLatestRun)
+  const latestRun = useSelector(selectLatestDryRun)
```
ramfox, line 37: could this be the source of the bug where only the dry run was showing updating the outline?

src/features/activityFeed/state/activityFeedActions.ts
```diff
@@ -14,15 +14,20 @@ export function loadDatasetLogs(qriRef: QriRef): ApiActionThunk {
...
       endpoint: `ds/activity`,
       method: 'POST',
       body: {
-        ref: `${qriRef.username}/${qriRef.name}`
+        ref: `${qriRef.username}/${qriRef.name}`,
+        pageSize: 100
+      },
+      query: {
+        pageSize: 100,
+        page: 2
```
ramfox: why are we fetching page 2?

src/features/workflow/state/workflowState.ts
```diff
@@ -30,6 +30,14 @@ export const SET_RUN_MODE = 'SET_RUN_MODE'
 // of the workflow without having to have a working api
 export const TEMP_SET_WORKFLOW_EVENTS = 'TEMP_SET_WORKFLOW_EVENTS'
 
+export const selectLatestDryRun = (state: RootState): Run | undefined => {
+  if (state.workflow.lastDryRunID) {
+    // console.log('calculating event log for id', state.workflow.lastRunID, 'from events', state.workflow.events, NewRunFromEventLog(state.workflow.lastRunID, state.workflow.events))
```
ramfox: stray log

src/features/workflow/state/workflowState.ts
```diff
@@ -69,6 +77,7 @@ export interface WorkflowState {
   // with working state to determine isDirty
   workflowBase: WorkflowBase
   isDirty: boolean
+  lastDryRunID?: string,
```
ramfox, line 78: why do we need to keep try of both the last dry run and the last run? once one or the other happens, wouldn't they wipe out any previous ui anyway? probably missing something, but want to flag it!

src/features/workflow/Workflow.tsx
chriswhong: I'm having trouble reproducing hat error, so it may be a fluke.

src/features/activityFeed/state/activityFeedActions.ts
chriswhong: Ah, I meant to remove all of this, I was trying to get more than 25 results from this endpoint and left it in there.

ramfox (approved):
think there might be some crossed wires surrounding fetching the dryrun from state and the deploy run from state, that may be causing that bug from this morning (where the outline wasn't showing the deploy run)

once that's cleaned up, LGTM!

src/features/workflow/state/workflowState.ts
chriswhong: One is used in the Run log to determine whether the Run now should be disabled, the other is used in the workfloweditor.  Assuming two runs can be happening simultaneously, we don't want them messing with each other's UI

PR #247, opened 2021-08-12
Some cleanup

chriswhong:
- cleans up all of the eslint warnings that were cluttering the console
- `API_DEPLOY_SUCCESS` was modifying a deployment's status and should not have been
- fixes some config that was causing warnings in `react-data-table-component`
- Fixed: New Workflow route was calling for data as if it were a dataset's workflow.


ramfox (approved):
LGTM!

PR #248, opened 2021-08-13
Various Improvements and Fixes

chriswhong:
Cleanup:

- Moves `ManualRunButton` out of its own feature and into `workflow`, gets rid of a lonely `manualRunButtonActions` and uses `runNow()`
- Cleans up `Icon`, making sure all icons used throughout the app are valid and exist, adds a warning if the component is used with an invalid icon name. 
- Refactor `selector` key in `react-data-table-component` column config, eliminates warnings and simplifies sorting
- Make `mini` prop optional in NewDatasetButton
- remove an unused import in DatasetWrapper
- clean up scrolling awkwardness in collection view
- get rid of the artificial delay on run toasts

Fixes: 

- fix: restore dry run keyboard shortcut which was broken because the action it called requires new arguments
- fix: restore sort by name in collection which was broken when we fixed warnings in an earlier commit
- fix: remove dataset from collection - app was crashing after remove 
- fix: only show LogItems with a runStatus in Run Log (Run Log was showing manual commits)
- fix: don't show number in sidenav icon if it is zero (manually updated datasets will have 0 runs)
- fix: add resetWorkflowState action, clear state when user navigates to `/workflow/new`


src/features/collection/CollectionTable.tsx
```diff
@@ -206,7 +206,7 @@ const CollectionTable: React.FC<CollectionTableProps> = ({
       omit: simplified,
       width: '90px',
       cell: (row: VersionInfo) => (row.workflowID
-          ? <ManualTriggerButton workflowID={row.workflowID} />
+          ? <ManualTriggerButton qriRef={qriRefFromString(`${row.username}/${row.name}`)} />
```
ramfox: This feels like one more step than needed

```
<ManualTriggerButton qriRef={{username: row.username, name: row.name}} />
```

src/features/collection/CollectionTable.tsx
```diff
@@ -79,9 +79,9 @@ export const customStyles = {
 
   },
   tableWrapper: {
-    style: {
-      display: 'flex'
-    },
+    // style: {
+    //   display: 'flex'
+    // },
```
ramfox: If these are commented out, but going to be commented back in in the future, please add a comment about when they should be added back in. Otherwise, probs should just remove the lines.

ramfox (approved):
LGTM, two small nits!

PR #250, opened 2021-08-19
feat: prevent endless loop on search, small layout fix

chriswhong:
- fixes and endless loop (and endless api calls) when loading `/search`
- fixes a layout bug that was causing the sort dropdown to appear in the wrong place


b5 (approved):
LGTM

PR #251, opened 2021-08-20
Refactor readme

chriswhong:
- Refactors `Readme` to take a `data` prop like other components
- Eliminates extra api call to `/render` on preview and history views, updates `Readme` type to use `text` instead of `script`
- Adds a client-side markdown renderer
- Fix readme layout on previews which was overflowing its container


src/qri/dataset.ts
```diff
@@ -404,14 +404,14 @@ export function scriptFromTransform(t: Transform): string {
 
 export interface Readme extends Component {
   scriptPath: string
-  script: string
+  text: string
 }
```
b5, line 408: 👍 backend folks really need to start warning FE teams about breaking changes like this

b5 (approved):

PR #252, opened 2021-08-20
fix: restore user profile, small layout tweaks

chriswhong:
- fixes the endless loop/api call bug when loading a user profile
- fixes a layout bug

### DIFF
--- a/src/features/userProfile/UserProfile.tsx
+++ b/src/features/userProfile/UserProfile.tsx
@@ -48,18 +48,18 @@ const UserProfile: React.FC<UserProfileProps> = ({ path = '/' }) => {
   const paginatedFollowingResults = useSelector(selectUserProfileFollowing)
 
   const scrollContainer = useRef<HTMLDivElement>(null)
-  const { username: usernameParam } = useParams();
+  const { username: usernameParam } = useParams()
 
   const { search } = useLocation()
   const userProfileParams: UserProfileDatasetListParams = NewUserProfileDatasetListParams(queryString.parse(search))
-  const { page, sort } = userProfileParams
 
   // if the query string ever changes, fetch new data
   useEffect(() => {
     dispatch(loadUserProfile(usernameParam))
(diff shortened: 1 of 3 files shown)


src/features/userProfile/UserProfile.tsx
```diff
@@ -59,7 +59,7 @@ const UserProfile: React.FC<UserProfileProps> = ({ path = '/' }) => {
     dispatch(loadUserProfile(usernameParam))
     dispatch(loadUserProfileDatasets(usernameParam, userProfileParams))
     dispatch(loadUserProfileFollowing(usernameParam, userProfileParams))
-  }, [usernameParam, page, sort, dispatch, userProfileParams])
+  }, [usernameParam, page, sort])
```
b5: this isn't the right fix. The thing causing an endless loop is on line 54 above:

```ts
  const userProfileParams: UserProfileDatasetListParams = NewUserProfileDatasetListParams(queryString.parse(search))
```

On that line we're creating a new `userProfileParams` object on every render, and passing that variable as a memoization argument to `useEffect`, which is telling `useEffect` it needs needs to be re-execute (and re-fetch) on every render, because one if it's arguments is always new.

Looking at that line above, the thing that we need to memoize on is `search`, which a string coming from the `useLocation` hook:

```ts
  useEffect(() => {
    const userProfileParams: UserProfileDatasetListParams = NewUserProfileDatasetListParams(queryString.parse(search))
     dispatch(loadUserProfile(usernameParam))
     dispatch(loadUserProfileDatasets(usernameParam, userProfileParams))
     dispatch(loadUserProfileFollowing(usernameParam, userProfileParams))
    }, [usernameParam, page, sort, dispatch, search])
```
search won't change every time, and our effect hook will be properly memoized. We'll need to construct a second `userProfileParams` for the render function below, but that's ok b/c the whole point of useEffect is sparse calling

b5 (approved):

PR #253, opened 2021-08-20
refactor: eliminate TextLink, add Link

chriswhong:
Previously we had `TextLink` and `ExternalLink`, causing some confusion.
`Link` now does it all with the following props:
```
  to?: string
  className?: string,
  colorClassName?: string,
  onClick?: () => void
```
If `to` exists, it will look for "http" in the string and render an `a` tag with `target="_blank"` etc for external links, or a react-router `Link` for internal links.

if `onClick` exists it will simply render a div but the classes will make text appear/behave like a hyperlink.

`Link` is then used to simplify in `DropdownMenuItem` to get consistent behavior for the various types of DropdownMenuItems (internal links, external links, onClicks)

`ExternalLink` and `TextLink` are replaced with `Link` throughout the app


src/chrome/DropdownMenuItem.tsx
```diff
@@ -24,22 +27,31 @@ const DropdownMenuItem: React.FC<DropdownMenuItemProps> = ({
...
+    'hover:cursor-pointer text-qrigray-400 hover:bg-gray-100': !disabled,
+    'text-qrigray-300': disabled
+  })
+
+  const colorClassName = 'text-qrigray-400'
+
+  const content = (
+    <div className={classNames({'text-qripink font-semibold': active })}>
+      {icon && <Icon icon={icon} className='mr-2' size='sm' />}{label}
+    </div>
```
b5: no need for `content` const. Inline directly into returned react element

src/chrome/DropdownMenuItem.tsx
```diff
@@ -2,20 +2,23 @@ import React from 'react'
...
 import Icon from './Icon'
+import Link from './Link'
 
 export const dropdownBaseClass = 'text-left text-xs px-2 py-0.5 mx-2 my-1 whitespace-nowrap rounded-md'
 
 export interface DropdownMenuItemProps {
   active?: boolean
   // label is the text that will be displayed to the user
-  label: string | React.ReactElement
+  label: string
```
b5, line 12: yay for string only!

src/chrome/Link.tsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react'
+import classNames from 'classnames'
+import { Link } from 'react-router-dom'
```
b5: use `as RouterLink` so you can get the `Link` name back:

```ts
import { Link as RouterLink } from 'react-router-dom`
```

rename `CustomLink` -> `Link`

src/chrome/Link.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const defaultColorClassName = 'text-qriblue hover:text-qriblue-800'
```
b5: no need for a `const` here, declare `'text-qriblue hover:text-qriblue-800'` directly as a default on the `Link` signature

src/chrome/Link.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const combinedClassNames = classNames('cursor-pointer transition-all duration-100', colorClassName)
+
+  // use a div by default
+  let theLink = (
+    <span onClick={onClick} className={combinedClassNames}>
+      {children}
+    </span>
+  )
+
+  // if to exists, make it a gatsby <Link>
```
b5: pretty sure this isn't a gatsby link 😄

b5 (approved):
Lovely change. a few nits, then LGTM

PR #281, opened 2021-08-24
feat: use read-only monaco editor to display transform script with syntax highlighting

chriswhong:
Achieves consistent display of transforms in the history viewer and workflow editor.

Home___Qri_Cloud


b5 (approved):
LGTM

PR #284, opened 2021-08-24
feat: fix runstatus tooltip z-index bug

chriswhong:
feat: fix runstatus tooltip z-index bug (makes sure tooltip is always in front)
fix: truncate name column in collection view (makes it so user cannot scroll sideways in collection view)


b5 (approved):
LGTM

PR #292, opened 2021-08-25
fix: multiple ui bugs

### LINKED ISSUE #289: UI Punchlist
- [x] Trigger: entire box shows counter cursor, but only “Edit” is clickable
- [x] User should not be able to dismiss deploy modal
- [x] Webhook trigger was clickable (it’s grayed out)

chriswhong:
Three fixes for issues found in user testing, closes #289 

- Trigger: entire box shows counter cursor, but only “Edit” is clickable
- User should not be able to dismiss deploy modal
- Webhook trigger was clickable (it’s grayed out)

### DIFF
--- a/src/features/app/modal/Modal.tsx
+++ b/src/features/app/modal/Modal.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useRef } from 'react'
+import React, { useCallback, useEffect, useRef, useState } from 'react'
 import { useDispatch, useSelector } from 'react-redux'
 
 import { ModalType, selectModal } from '../state/appState'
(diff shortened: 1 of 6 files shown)


src/features/app/modal/Modal.tsx
```diff
@@ -19,21 +19,31 @@ const Modal: React.FC<any> = () => {
   const dispatch = useDispatch()
   const modal = useSelector(selectModal)
 
+
+
+  // this is called from from multiple places (esc key press or click outside modal)
+  // and is where we finally check allowClear and
```
b5: unfinished comment. I'd recommend removing entirely

src/features/app/state/appActions.ts
```diff
@@ -1,10 +1,15 @@
...
 
 export interface ModalAction {
   type: string
   modal: Modal
 }
 
+export interface ModalLockedAction {
+  type: string
+  locked: boolean
+}
```
b5, line 12: in addition to the  `SetModalLocked` action, add a `locked` property to `ModalAction`:

```ts
export interface ModalAction {
  type: string
  modal: Modal
  locked?: bool
}
```

and add a new optional `locked` arg to `showModal` that defaults to false

```ts
export function showModal<P = {}>(type: ModalType, props?: P, locked?: bool = false): ModalAction
```

we can't make use of this quite yet thanks to the way the deploy modal is designed, but we can make use of it in a follow up

b5 (approved):
one request for an API change, otherwise looks good!

PR #296, opened 2021-08-25
fix: reset isDirty after fetching preview and workflow

### LINKED ISSUE #280: isDirty is not reset between new and existing workflow
The workflow state is not being properly reset:
- New Workflow
- make an edit
- leave, clicking discard changes
- go to existing dataset with workflow, click workflow
- button will be enabled when there are no changes

chriswhong:
- Ensures that `workflowBase` is properly updated after fetching dataset preview and workflow, and that `isDirty` is false when navigating to a dataset.

- Cleans up some typescript errors

Closes #280

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "@testing-library/react": "^9.3.2",
     "@testing-library/user-event": "^7.1.2",
     "@types/classnames": "^2.2.11",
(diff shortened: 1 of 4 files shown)


b5 (approved):
LGTM

PR #279, opened 2021-08-24
Carousel arrows

chriswhong:
- Add arrows to workflow splash carousel
- Don't show "New Dataset" button in sidebar when on /workflow/new route
- Loading spinner for history>body pane 
- Remove DatasetNavBar (unused)
- Fix Short body table appears vertically centered

PR #298, opened 2021-08-25
refactor: Componentize DatasetCommitInfo

chriswhong:
Builds on work in #279 which should be merged first

Adds a new component `DatasetCommitInfo`, providing a two-row layout with a commit's title, automation status, user, relative timestamp, and commit.

Some version of this commit summary layout was in place in preview, multiple places in history, and in the run log when a run results in a commit.  This standardizes the layout, ensuring that the commit info looks consistent across all of these locations in the app.

Home___Qri_Cloud

### DIFF
--- a/src/chrome/DatasetCommitInfo.tsx
+++ b/src/chrome/DatasetCommitInfo.tsx
@@ -1,0 +1,45 @@
+import React from 'react'
+import classNames from 'classnames'
+
+import Icon from './Icon'
+import RelativeTimestampWithIcon from './RelativeTimestampWithIcon'
+import UsernameWithIcon from './UsernameWithIcon'
+import commitishFromPath from '../utils/commitishFromPath'
+import { Dataset } from '../qri/dataset'
+
+interface DatasetCommitInfoProps {
+  dataset: Dataset
+  // small yields the same info with smaller text, used in history list and run log
+  small?: boolean
+}
+
+const DatasetCommitInfo: React.FC<DatasetCommitInfoProps> = ({
+  dataset,
+  small=false
+}) => {
+  return (
+    <div className={classNames({
+      'text-sm': !small,
+      'text-xs': small
+    })}>
+      <div className={classNames('text-qrinavy font-semibold flex justify-between items-center mb-2')}>
+        <div>{dataset.commit?.title}</div>
+        <div className='flex-grow-0 text-qrigreen'>
+          <Icon icon='automationFilled' size={small ? 'xs' : 'sm'}/>
+        </div>
+      </div>
+      <div className='flex items-center text-gray-400'>
+        <UsernameWithIcon username={dataset.username} className='mr-2' iconWidth={small ? 14 : 18} iconOnly={small} />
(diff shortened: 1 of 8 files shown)


src/chrome/DatasetCommitInfo.tsx
```diff
@@ -0,0 +1,49 @@
+// DatasetCommitInfo is a standardized layout for info about a version
+// useful for showing the latest commit in previews, the selected commit in history,
+// lists of commits, etc
```
b5: it's tempting to use phrases like "standardized" when doing this sort of refactoring. Think about re-reading this comment in six months. What does standardized mean in this context? I don't think the comment is really necessary. Instead the fact that it's used in multiple places should speak to it's standardization

src/chrome/DatasetCommitInfo.tsx
```diff
@@ -0,0 +1,49 @@
...
+  dataset,
+  small=false
+}) => {
+  return (
+    <div className={classNames({
+      'text-sm': !small,
+      'text-xs': small
+    })}>
+      <div className={classNames('text-qrinavy font-semibold flex justify-between items-center mb-2')}>
+        <div className=''>{dataset.commit?.title}</div>
```
b5: remove empty `className`

src/chrome/DatasetCommitInfo.tsx
```diff
@@ -0,0 +1,49 @@
...
+          <Icon icon='automationFilled' size={small ? 'xs' : 'sm'}/>
+        </div>
+      </div>
+      <div className='flex items-center text-gray-400'>
+        <UsernameWithIcon username={dataset.username} className='mr-2' iconWidth={small ? 14 : 18} iconOnly={small} />
+        <RelativeTimestampWithIcon className='mr-3' timestamp={new Date(dataset.commit?.timestamp)} />
+        {dataset.path && (
+          <div className='flex items-center leading-tight'>
+            <Icon icon='commit' size={small ? 'xs' : 'sm'} className='-ml-2' />
+            <div>{commitishFromPath(dataset.path)}</div>
```
b5, line 37: at some point I really need a tooltip here that shows the full commit, or a requirement that these commitish's are always links. I often try to copy hashes from here to go search IPFS, and get frustrated that I can't

src/features/commits/DatasetCommit.tsx
```diff
@@ -1,15 +1,15 @@
...
-import { Link } from 'react-router-dom';
-import classNames from 'classnames';
+import { Link } from 'react-router-dom'
+import classNames from 'classnames'
+
+import { LogItem } from '../../qri/log'
+import { newQriRef } from '../../qri/ref'
+import { NewDataset } from '../../qri/dataset'
+import { pathToDatasetViewer } from '../dataset/state/datasetPaths'
+// import ComponentChangeIndicatorGroup from './ComponentChangeIndicatorGroup'
```
b5: remove ded import

b5 (approved):

src/chrome/DatasetCommitInfo.tsx
chriswhong: I'll add an issue for this!

PR #302, opened 2021-08-26
fix: fixed issue of user menu z-index on new datasets page

boandriy:
fix: fixed issue of user menu z-index on new datasets page, where user menu popup overlapped new dataset from workflow menu on scroll.

After fix: 
image


.gitignore
```diff
@@ -18,6 +18,7 @@
 .env.test.local
 .env.production.local
 .eslintcache
+.idea
```
boandriy, line 21: I am working in Webstorm. It creates the .idea folder, so added it here.

b5 (approved):
Looks great!

PR #307, opened 2021-08-26
fix: handle no ref on apply

chriswhong:
- handles no `qriRef` in the `applyWorkflowTransform` restoring the ability to dry run
- adds `wait: false` to the `applyWorkflowTransform` api call so that the response comes back immediately and the step-based `RunStatus` indicators function properly

### DIFF
--- a/src/features/workflow/state/workflowActions.ts
+++ b/src/features/workflow/state/workflowActions.ts
@@ -113,11 +113,11 @@ export function applyWorkflowTransform(w: Workflow, d: Dataset): ApiActionThunk 
         endpoint: 'auto/apply',
         method: 'POST',
         body: {
-          ref: `${qriRef.username}/${qriRef.name}`,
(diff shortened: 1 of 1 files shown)


src/features/workflow/state/workflowActions.ts
```diff
@@ -113,11 +113,11 @@ export function applyWorkflowTransform(w: Workflow, d: Dataset): ApiActionThunk
         endpoint: 'auto/apply',
         method: 'POST',
         body: {
-          ref: `${qriRef.username}/${qriRef.name}`,
-          wait: true,
+          ref: (qriRef.username && qriRef.name) ? `${qriRef.username}/${qriRef.name}` : '',
```
b5, line 116: there is a better fix for this that involves turning qriRef into a concrete class, but this works for now

b5 (approved):

PR #322, opened 2021-09-10
refactor: extended versionInfo interface…

### LINKED ISSUE #217: Info items in DatasetHeader are still showing dummy data
Need `stats` in the `/get` response, qri core issue: https://github.com/qri-io/qri/issues/1823

### LINKED ISSUE #312: refactor(): adjust `VersionInfo` to temporarily add aggregate fields
We have some missing data in the the `DatasetHeader`, as described by issues #259, #217, and qri-io/qri#1823 & qri-io/qri#1880

Comment https://github.com/qri-io/qri/issues/1880#issuecomment-910422967 describes how we are going to solve this for now in the backend. 

For the frontend, we also need to make adjustments to the `VersionInfo` type, and need to ensure that we have comments/TODOs that these fields will likely not exist on `VersionInfo` long term.

We also need to add an action for getting these `VersionInfo`s individually by fetching from the `collection/get` endpoint, as well as ensuring they end up in the state tree, and that we fetch them at the correct lifecycle.

#### adjusting VersionInfo type
**note, we are renaming numVersions to commitCount**
That will need to adjusted elsewhere in the app.

```typescript
// src/qri/versionInfo.ts

export interface VersionInfo {
...
  //
  // aggregation details
  // TODO (ramfox): These fields are only temporarily living on `VersionInfo`.
  // When we get more user feedback and settle what info
  // users want about their datasets, these fields may move to a new struct
  // store, or subsystem. 
  // These fields are not derived from any `dataset.Dataset` fields
  runCount?: number        // number of time this dataset's transform has run
  commitCount?: number       // number of commits in this dataset's history
  downloadCount?: number        // number of times this dataset has been directly downloaded
  followerCount?: number // number of followers this dataset has
  openIssueCount?: number // number of open issues this dataset has
}

export function newVersionInfo(data: Record<string,any>): VersionInfo {
  return {
    ...
    // TODO(ramfox): the following fields are likely to be removed from the VersionInfo
    // type in the future
    runCount: data.runCount,
    commitCount: data.commitCount,
    downloadCount: data.downloadCount,
    followerCount: data.followerCount,
    openIssueCount: data.openIssueCount
  }
}

```

#### Adding `fetchHeader` action
```typescript
// src/features/datasets/state/datasetActions

func function mapVersionInfo(d: object | []): VersionInfo {
	return NewVersionInfo((d as Record<string,any))
}

// fetchHeader fetches a specified `VersionInfo` from the collection
// TODO(ramfox): this endpoint ('collection/get') returns a `VersionInfo`, but
// this return type may change in the future when the specific "dataset details"
// that a user wants to see on the dataset page settles. We may then have a new
// data structure that returns from the 'collection/get' endpoint
export func fetchHeader (ref: QriRef): ApiAction {
	return {
		type: 'header',
		ref,
		[CALL_API]: {
			endpoint: 'collection/get',
			method: 'GET',
			segments: ref,
			map: mapVersionInfo,
		}
	}
}
```

#### add `header` to `dataset` page state tree, add header action types to reducer, and add `selectHeader`
I'm leaving out header loading state details that should be fleshed out
```typescript
// src/features/datasets/state/datasetState.ts

export const selectHeader = (state: RootState): VersionInfo => state.dataset.header

export interface DatasetState {
  dataset: Dataset
  // TODO(ramfox): header is currently a `VersionInfo`, but this is likely to change to its own data structure in the future
  header: VersionInfo
  loading: boolean
}

const initialState: DatasetState = {
  dataset: NewDataset({}),
  header: NewVersionInfo({}),
  loading: true
}

export const datasetReducer = createReducer(initialState, {
  'API_HEADER_REQUEST': (state) => {
   state.header = NewVersionInfo({})
  },
  'API_HEADER_SUCCESS': (state) => {
    state.header = action.payload.data as VersionInfo
  },
  'API_HEADER_FAILURE': (state) => {
    // loading state
  },
 })
```

#### fetch the dataset header
The `DatasetComponents` component should be modified to fetch and a `header` if no header currently exists or if the `QriRef` changes. 
The `DatasetFixedLayout` should be modified to select the `header` from state using `selectHeader`

boandriy:
replaced the dataset header/sidebar prop numbers with ones that come from BE.
extended the versionInfo interface to include : 
`RunCount,
  CommitCount,
  DownloadCount,
  FollowerCount,
  OpenIssueCount`

### DIFF
--- a/src/features/dataset/DatasetHeader.tsx
+++ b/src/features/dataset/DatasetHeader.tsx
@@ -1,18 +1,19 @@
 import React from 'react'
-import { useDispatch } from 'react-redux'
+import { useDispatch, useSelector } from 'react-redux'
 import { useHistory } from 'react-router'
 
 import EditableLabel from '../../chrome/EditableLabel'
 import { renameDataset } from './state/datasetActions'
-import { Dataset, qriRefFromDataset } from '../../qri/dataset'
 import DatasetInfoItem from './DatasetInfoItem'
 import DownloadDatasetButton from '../download/DownloadDatasetButton'
 import Link from '../../chrome/Link'
 import { validateDatasetName } from '../session/state/formValidation'
+import { selectDatasetHeader } from "./state/datasetState";
+import { qriRefFromVersionInfo } from "../../qri/versionInfo";
+import fileSize from "../../utils/fileSize";
 
 
 export interface DatasetHeaderProps {
-  dataset: Dataset
   border?: boolean
   editable?: boolean
   showInfo?: boolean
@@ -23,7 +24,6 @@ export interface DatasetHeaderProps {
 // overriding the download button with the dry run button in the workflow editor
 
 const DatasetHeader: React.FC<DatasetHeaderProps> = ({
-  dataset,
   border = false,
   editable = false,
   showInfo = true,
@@ -31,13 +31,15 @@ const DatasetHeader: React.FC<DatasetHeaderProps> = ({
 }) => {
   const dispatch = useDispatch()
   const history = useHistory()
-  const qriRef = qriRefFromDataset(dataset)
+  const header = useSelector(selectDatasetHeader)
+  const qriRef = qriRefFromVersionInfo(header)
+
 
   const handleRename = (_:string, value:string) => {
-    renameDataset(qriRef, { username: dataset.peername, name: value })(dispatch)
+    renameDataset(qriRef, { username: header.username, name: value })(dispatch)
       .then(({ type }) => {
         if (type === 'API_RENAME_SUCCESS') {
-          const newPath = history.location.pathname.replace(dataset.name, value)
+          const newPath = history.location.pathname.replace(header.name, value)
           history.replace(newPath)
         }
       })
@@ -62,15 +64,15 @@ const DatasetHeader: React.FC<DatasetHeaderProps> = ({
           )}
 
           <div className='text-2xl text-qrinavy-500 font-black group hover:text'>
-            {dataset.meta?.title || dataset.name}
+            {header.name}
           </div>
           {showInfo && (
             <div className='flex mt-3 text-sm'>
-              <DatasetInfoItem icon='automationFilled' label='automated' iconClassName='text-qrigreen' />
-              <DatasetInfoItem icon='disk' label='59 MB' />
-              <DatasetInfoItem icon='download' label='418 downloads' />
-              <DatasetInfoItem icon='follower' label='130 followers' />
-              <DatasetInfoItem icon='lock' label='private' />
+              {header.workflowID && <DatasetInfoItem icon='automationFilled' label='automated' iconClassName='text-qrigreen' />}
+              <DatasetInfoItem icon='disk' label={fileSize(header.bodySize || 0)} />
+              <DatasetInfoItem icon='download' label={getLabel(header.downloadCount, 'download')} />
+              <DatasetInfoItem icon='follower' label={getLabel(header.followerCount, 'follower')} />
+              <DatasetInfoItem icon='globe' label='public' />
             </div>
           )}
         </div>
@@ -98,3 +100,7 @@ const DatasetHeader: React.FC<DatasetHeaderProps> = ({
 }
 
 export default DatasetHeader;
+
+function getLabel(count: number, label: string) {
+  return `${count} ${label}${count === 1 ? '' : 's'}`
+}
--- a/src/features/dataset/DatasetRoutes.tsx
+++ b/src/features/dataset/DatasetRoutes.tsx
@@ -5,8 +5,9 @@
 
 // DatasetPreviewPage fetches the other necessary parts of the preview (body + readme)
 
-import React from 'react'
+import React, {useEffect} from 'react'
 import { Redirect, Route, Switch, useParams, useRouteMatch } from 'react-router'
+import { useDispatch } from "react-redux";
 
 import WorkflowPage from '../workflow/WorkflowPage'
 import DatasetComponents from '../dsComponents/DatasetComponents'
@@ -16,6 +17,7 @@ import DatasetIssues from '../issues/DatasetIssues'
 import { newQriRef } from '../../qri/ref'
 import DatasetEditor from '../dsComponents/DatasetEditor'
 import DatasetWrapper from '../dsComponents/DatasetWrapper'
+import { loadHeader } from "./state/datasetActions";
 
 const DatasetRoutes: React.FC<{}> = () => {
   const { url } = useRouteMatch()
@@ -26,6 +28,11 @@ const DatasetRoutes: React.FC<{}> = () => {
   // ref. Move all ref constructino into container components to avoid potential
   // bugs
   const qriRef = newQriRef(useParams())
+  const dispatch = useDispatch()
+
+  useEffect(() => {
+    dispatch(loadHeader(qriRef))
+  },[dispatch, qriRef.username, qriRef.name, qriRef.path]);
 
   return (
     <DatasetWrapper>
--- a/src/features/dataset/state/datasetActions.ts
+++ b/src/features/dataset/state/datasetActions.ts
@@ -1,8 +1,9 @@
 import Dataset, { NewDataset } from "../../../qri/dataset"
 import { QriRef, refStringFromQriRef, humanRef } from "../../../qri/ref"
 import { ApiAction, ApiActionThunk, CALL_API } from "../../../store/api"
-import { RENAME_NEW_DATASET } from "./datasetState"
+import {RENAME_NEW_DATASET, SET_HEADER} from "./datasetState"
 import { API_BASE_URL } from '../../../store/api'
+import {VersionInfo} from "../../../qri/versionInfo";
 
 export const bodyPageSizeDefault = 50
 
@@ -17,6 +18,27 @@ export function loadDataset(ref: QriRef): ApiActionThunk {
   }
 }
 
+export function loadHeader(ref: QriRef): ApiActionThunk {
+  return async (dispatch, getState) => {
+    if(getState){
+      const state = getState()
+      const existingHeader: VersionInfo | undefined = getExistingHeader(state.collection.collection, ref)
+      if(existingHeader)
+        return dispatch(setHeader(existingHeader))
+    }
+    return dispatch(fetchHeader(ref))
+  }
+}
+
+function getExistingHeader(collection: Record<string, VersionInfo>, ref: QriRef): VersionInfo | undefined {
+  for (const k in collection) {
+    if(collection[k].name === ref.name && collection[k].username === ref.username){
+      return collection[k]
+    }
+  }
+  return undefined
+}
+
 // downloadLinkFromQriRef creates a download link
 export function downloadLinkFromQriRef(ref: QriRef, body: boolean = false): string {
   let pathSegment = ref.path ? `/at${ref.path}` : ''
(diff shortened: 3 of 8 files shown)


b5:
looks like this is related to #217.

src/qri/versionInfo.ts
```diff
@@ -36,6 +36,12 @@ export interface VersionInfo {
   runID?: string
   runStatus?: RunStatus
   runDuration?: number
+
+  runCount?: number // RunCount is the number of times this dataset's transform has been run
+  commitCount?: number // CommitCount is the number of commits in this dataset's history
+  downloadCount?: number // DownloadCount is the number of times this dataset has been directly downloaded from this Qri node
+  followerCount?: number // FollowerCount is the number of followers this dataset has on this Qri node
+  openIssueCount?: number // OpenIssueCount is the number of open issues this dataset has on this Qri node
```
b5: I'd much rather follow the golang convention of using a default value for the number instead of making these fields optional. To accomplish that in typescript make all of these fields required (remove the `?`), ensure all versionInfo instantiation passes through the `newVersionInfo` constructor, and add `|| 0` to the end of each value assignment. Eg: `runCount: data.runCount || 0`.

This removes the need for existential checks all of the place. It won't be true for all verionInfo fields yet, but hey, gotta start somewhere.

src/features/dataset/state/datasetState.ts
```diff
@@ -27,15 +34,23 @@ export const selectSessionUserCanEditDataset = (state: RootState): boolean => {
 
 export interface DatasetState {
   dataset: Dataset
+  datasetVersionInfo: VersionInfo
```
b5: this can just be `versionInfo`

src/features/dataset/DatasetRoutes.tsx
```diff
@@ -26,6 +28,12 @@ const DatasetRoutes: React.FC<{}> = () => {
   // ref. Move all ref constructino into container components to avoid potential
   // bugs
   const qriRef = newQriRef(useParams())
+  const dispatch = useDispatch()
+
+  useEffect(() => {
+    dispatch(loadDatasetVersionInfo(qriRef))
+    //eslint-disable-next-line
```
b5: is the line below complaining about an empty array? it'd be better to remove the array entirely and avoid the eslint disable.

src/features/dataset/DatasetRoutes.tsx
```diff
@@ -16,6 +16,8 @@ import DatasetIssues from '../issues/DatasetIssues'
 import { newQriRef } from '../../qri/ref'
 import DatasetEditor from '../dsComponents/DatasetEditor'
 import DatasetWrapper from '../dsComponents/DatasetWrapper'
+import { useDispatch } from "react-redux";
```
b5: organize imports in the general form:
```
(external imports / modules)
BLANK LINE
(relative imports)
```

src/features/dataset/DatasetHeader.tsx
```diff
@@ -68,8 +70,8 @@ const DatasetHeader: React.FC<DatasetHeaderProps> = ({
             <div className='flex mt-3 text-sm'>
               <DatasetInfoItem icon='automationFilled' label='automated' iconClassName='text-qrigreen' />
               <DatasetInfoItem icon='disk' label='59 MB' />
-              <DatasetInfoItem icon='download' label='418 downloads' />
-              <DatasetInfoItem icon='follower' label='130 followers' />
+              <DatasetInfoItem icon='download' label={`${versionInfo?.downloadCount || 0} downloads`} />
+              <DatasetInfoItem icon='follower' label={`${versionInfo?.followerCount || 0} followers`} />
```
b5: refactoring versionInfo to always have numeric fields set makes it much easier to ship pluralization. Define label funcs at the bottom of this file:

```ts
function downloadLabel(vi?: VersionInfo) string {
  if (!vi) { return `0 downloads` }
  const verb = (vi.downloadCount === 1) ? 'download' : 'downloads'
  return `${vi.downloadCount} ${verb}`
}
```
```suggestion
              <DatasetInfoItem icon='download' label={downloadCount(versionInfo)} />
```

ditto for `followerCount`

src/qri/versionInfo.ts
```diff
@@ -36,6 +36,12 @@ export interface VersionInfo {
   runID?: string
   runStatus?: RunStatus
   runDuration?: number
+
```
ramfox, line 39: need to add a todo here saying that these fields are temporary.

For example:
```
  // TODO (boandriy): These fields are only temporarily living on `VersionInfo`.
  // When we get more user feedback and settle what info
  // users want about their datasets, these fields may move to a new struct
  // store, or subsystem. 
```

src/qri/versionInfo.ts
```diff
@@ -69,6 +75,13 @@ export function newVersionInfo(data: Record<string,any>): VersionInfo {
     runID: data.runID,
     runStatus: data.runStatus,
     runDuration: data.runDuration,
+
```
ramfox, line 82: again, add a TODO here mentioning the fields as temporary additions:
```
// TODO(boandriy): the following fields are likely to be removed from the VersionInfo
    // type in the future
```

src/features/dataset/state/datasetActions.ts
```diff
@@ -27,6 +33,19 @@ export function downloadLinkFromQriRef(ref: QriRef, body: boolean = false): stri
   return `${API_BASE_URL}/ds/get/${ref.username}/${ref.name}${pathSegment}?format=zip`
 }
 
+function fetchDatasetVersionInfo(ref: QriRef) {
```
ramfox: As per #312, we should call this function `fetchHeader` & we need to add a TODO here explaining that the `collection/get` endpoint is only temporarily returning `VersionInfos`:

```
// fetchHeader fetches a specified `VersionInfo` from the collection
// TODO(boandriy): this endpoint ('collection/get') returns a `VersionInfo`, but
// this return type may change in the future when the specific "dataset details"
// that a user wants to see on the dataset page settles. We may then have a new
// data structure that returns from the 'collection/get' endpoint
```

Again, because this endpoint will not always return `VersionInfo`, it would be inaccurate to call it `fetchDatasetVersionInfo`.

src/features/dataset/state/datasetActions.ts
```diff
@@ -17,6 +17,12 @@ export function loadDataset(ref: QriRef): ApiActionThunk {
   }
 }
 
+export function loadDatasetVersionInfo(ref: QriRef): ApiActionThunk {
```
ramfox: as below, this method should be called `loadHeader`. It is only temporarily returning `VersionInfo`s

src/features/dataset/state/datasetState.ts
```diff
@@ -5,11 +5,18 @@ import Dataset, { NewDataset } from '../../../qri/dataset';
 import { qriRefFromString } from '../../../qri/ref';
 import { RenameDatasetAction } from './datasetActions';
 import { selectSessionUser } from '../../session/state/sessionState';
+import { newVersionInfo, VersionInfo } from "../../../qri/versionInfo";
 
 export const RENAME_NEW_DATASET = 'RENAME_NEW_DATASET'
 
 export const selectDataset = (state: RootState): Dataset => state.dataset.dataset
 
+export const selectDatasetVersionInfo = (state: RootState): VersionInfo => state.dataset.datasetVersionInfo
```
ramfox: rename to `selectHeader`

src/features/dataset/state/datasetState.ts
```diff
@@ -5,11 +5,18 @@ import Dataset, { NewDataset } from '../../../qri/dataset';
...
+
+export const selectDatasetVersionCount = (state: RootState): number | undefined => state.dataset.datasetVersionInfo.commitCount
```
ramfox: rename to `selectCommitCount`

src/features/dataset/state/datasetState.ts
```diff
@@ -5,11 +5,18 @@ import Dataset, { NewDataset } from '../../../qri/dataset';
...
+
+export const selectDatasetLogCount = (state: RootState): number | undefined => state.dataset.datasetVersionInfo.commitCount
```
ramfox: this method seems to do the same thing as the above method. We should remove this and have any calls to this function be replaced by `selectCommitCount`

src/features/dataset/state/datasetState.ts
```diff
@@ -27,15 +34,23 @@ export const selectSessionUserCanEditDataset = (state: RootState): boolean => {
...
 }
 
 const initialState: DatasetState = {
   dataset: NewDataset({}),
+  datasetVersionInfo: newVersionInfo({}),
   loading: true
 }
 
 export const datasetReducer = createReducer(initialState, {
+  'API_VERSION_REQUEST': (state) => {
```
ramfox: should be `API_HEADER_REQUEST`

src/features/dataset/state/datasetState.ts
```diff
@@ -27,15 +34,23 @@ export const selectSessionUserCanEditDataset = (state: RootState): boolean => {
...
+    state.datasetVersionInfo = newVersionInfo({})
+  },
+  'API_VERSION_SUCCESS': (state, action) => {
```
ramfox: should be `API_HEADER_SUCCESS`

src/features/dataset/state/datasetState.ts
ramfox: this should actually be called `header` & we should add a note:
```
// TODO(boandriy): header is currently a `VersionInfo`, but will have its own data structure in future iterations
```

ramfox (changes requested):
Hey @boandriy, take another look at https://github.com/qri-io/frontend/issues/312

This PR should be closing that issue. 

In this current commit you are fetching the version info in `DatasetFixedLayout`, but it should be getting fetched in the `DatasetComponents` component. `DatasetFixedLayout` should select the necessary verison info from state, rather than fetching it itself.

src/features/dataset/DatasetHeader.tsx
```diff
@@ -62,14 +63,14 @@ const DatasetHeader: React.FC<DatasetHeaderProps> = ({
           )}
 
           <div className='text-2xl text-qrinavy-500 font-black group hover:text'>
-            {dataset.meta?.title || dataset.name}
+            {header.name}
           </div>
           {showInfo && (
             <div className='flex mt-3 text-sm'>
               <DatasetInfoItem icon='automationFilled' label='automated' iconClassName='text-qrigreen' />
               <DatasetInfoItem icon='disk' label='59 MB' />
```
ramfox: ahh, okay, so there are two pieces missing here, sorry I didn't catch this before:

1) we should only show the `DatasetInfoItem` that says "automated" if the version info has `workflowID` filled in
2) In the `DatasetInfoItem` with label '59 MB', that should be replaced by the actual size of the dataset taken from the version info, we also have a util function `fileSize` that you can import that displays the file size appropriately:
```js
<DatasetInfoItem icon='disk' label={`fileSize(header.bodySize)`} />
```

src/qri/versionInfo.ts
ramfox: hey hey, just pointing out this out!

src/features/dataset/DatasetHeader.tsx
```diff
@@ -62,14 +63,14 @@ const DatasetHeader: React.FC<DatasetHeaderProps> = ({
...
-              <DatasetInfoItem icon='download' label='418 downloads' />
-              <DatasetInfoItem icon='follower' label='130 followers' />
+              <DatasetInfoItem icon='download' label={getLabel(header.downloadCount, 'download')} />
+              <DatasetInfoItem icon='follower' label={getLabel(header.followerCount, 'follower')} />
               <DatasetInfoItem icon='lock' label='private' />
```
ramfox: Currently all datasets are public. Let's reflect that here. Rather than defaulting to 'private', let's default to 'public' & use the 'globe' icon.

```
<DatasetInfoItem icon='globe' label='public' />
```

ramfox (changes requested):
Just a few small fixes!

ramfox (approved):
Just add that comment in at https://github.com/qri-io/frontend/pull/322/files#r707581803

And this is good to merge!

PR #327, opened 2021-09-14
refactor: hidden the user name/description/socials

### LINKED ISSUE #320: Hide user profile header UI items that the user can't set
image

Social media links, name, description, etc aren't editable by users (yet).  We should hide them until we introduce the ability to edit them.

boandriy:
Hidden users description, users social links and username until there is an option to change them.

Fixes #320


chriswhong (changes requested):
When we leave TODO comments in code, we make sure to "own" them by adding the developer's username.  So, these should say `TODO(boandriy): ...`

Otherwise LGTM!

PR #329, opened 2021-09-15
fix(trigger): cron trigger must start sometime after first deploy

### LINKED ISSUE #317: bug: cron trigger timing is off
The way cron triggers work, if a trigger gets saved that has the start time as the current time or any time before the current time, it will immediately run the workflow when the trigger gets saved.

Since we want the ideal behavior to be that the run will only run immediately if the "run now" checkbox is filled, that means we need the trigger to have the proper "start time". So if the interval is "every ten minutes", then the trigger that we send over should have a start time ten minutes from now. 

This is currently causing double runs to happen on datasets if you add or update a trigger when "Deploying".

ramfox:
closes #317

### DIFF
--- a/src/features/trigger/util.ts
+++ b/src/features/trigger/util.ts
@@ -90,13 +90,15 @@ export const triggerFromSchedule = (schedule: Schedule): CronTrigger => {
   // for every 10 minutes, periodicity is 'R/PT10M'
   switch (schedule.periodicity) {
     case 'PT10M':
-      // get most recent previous hour, the ten-minutes interval will run at 00, 10, 20, 30, ...
+      const min = today.getMinutes()
+      const hr = min - 50 > 0 ? today.getHours() + 1 : today.getHours()
+      // start on the next closest 10 minute interval
       startTime = new Date(
         today.getFullYear(),
         today.getMonth(),
         today.getDate(),
-        today.getHours(),
-        0,
+        hr,
+        min + (10 - (min % 10)),
         0
       ).toISOString()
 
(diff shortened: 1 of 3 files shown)


src/features/trigger/util.ts
```diff
@@ -90,13 +91,15 @@ export const triggerFromSchedule = (schedule: Schedule): CronTrigger => {
...
+      const hr = min - 50 > 0 ? today.getHours() + 1 : today.getHours()
+      // start on the next closest 10 minute interval
       startTime = new Date(
         today.getFullYear(),
         today.getMonth(),
         today.getDate(),
-        today.getHours(),
-        0,
+        hr,
+        min + (10 - (min % 10)),
```
ramfox, line 102: would be great to get another eye on this logic @boandriy

src/features/trigger/util.ts
```diff
@@ -105,12 +108,12 @@ export const triggerFromSchedule = (schedule: Schedule): CronTrigger => {
...
       // for hourly, periodicity is R/P1H and startTime is the next instance of
       // the minutes selected by the user
-      // get most recent previous hour, add minutes, convert to ISO8601
+      const hrs = parseInt(schedule.minutes) <= today.getMinutes() ? today.getHours() + 1 : today.getHours()
       startTime = new Date(
         today.getFullYear(),
         today.getMonth(),
         today.getDate(),
-        today.getHours(),
+        hrs,
```
ramfox, line 116: again, would be great to get another person to look over this logic @boandriy

src/features/trigger/util.ts
```diff
@@ -120,12 +123,12 @@ export const triggerFromSchedule = (schedule: Schedule): CronTrigger => {
...
        // time is in HH:MM format in local time (24 hour hours)
       const [hours, minutes] = schedule.time.split(':')
-
+      const date = parseInt(hours) <= today.getHours() ? today.getDate() + 1 : today.getDate()
       // get ISO8601 string for today at the specified hours+minutes
       startTime = new Date(
         today.getFullYear(),
         today.getMonth(),
-        today.getDate(),
+        date,
```
ramfox, line 131: here as well @boandriy

boandriy (approved):
LGTM!

PR #330, opened 2021-09-16
fix: disabling the code cells when dry run is running

### LINKED ISSUE #291: Lock out workflow UI on dry run
During an interview, the user noted that the code blocks were still editable while the dry run was running.  We should lock out all of the UI on the workflow editor while dry run is executing, and unlock it again on success or failure.

boandriy:
disabled code cells on dry runs.

fixes #291


ramfox (approved):
This looks great! LGTM.

PR #331, opened 2021-09-16
fix: filtering out toasts for dry runs

### LINKED ISSUE #257: Do not show toasts for dry runs
Dry runs already happen on the workflow editor which has plenty of visual feedback that the job is running... no need to show a toast as well.

boandriy:
filtering out toasts for dry runs on workflow

fixes #257


ramfox (approved):
Just tested locally, works great!

LGTM.

PR #332, opened 2021-09-16
feat(Icon): add `Unlock` icon & use it in `DatasetHeader`

ramfox:
Also added instructions on how to import icons from the Figma board
to the `src/chrome/Icon.tsx` file


chriswhong (approved):
I think a yarn.lock change crept in here, FYI.  Everything else LGTM

PR #340, opened 2021-09-22
refactor: update default workflow template code

chriswhong:
image

Updates the default `template` for new workflows to include helpful comments and  a working transform that makes use of the new `dataframe` package in starlib (pulls baby names dataset, filters for `startsWith`, gets unique values, sorts, lowercases)

### DIFF
--- a/src/features/template/templates.ts
+++ b/src/features/template/templates.ts
@@ -34,19 +34,72 @@ export const CSVDownload: Dataset = {
   },
   transform: {
     steps: [
-      { syntax: 'starlark', category: 'setup', name: 'setup', script: `# load starlark dependencies
-load("http.star", "http")
-load("encoding/csv.star", "csv")` },
-      { syntax: 'starlark', category: 'download', name: 'download', script: `# get the popular baby names dataset as a csv
-def download(ctx):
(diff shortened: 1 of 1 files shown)


src/features/template/templates.ts
```diff
@@ -34,19 +34,73 @@ export const CSVDownload: Dataset = {
...
+`
+      },
+      {
+        syntax: 'starlark',
+        script: `# CSV Download Code Sample
+# This really works! Click 'Dry Run' to try it ↗
+
+# import Starlark dependencies
+load("http.star", "http") # \`http\` lets us talk to the internets
+load("encoding/csv.star", "csv") # \`csv\` lets us parse and output CSVs
```
b5: don't think we actually need the csv package for this script

src/features/template/templates.ts
```diff
@@ -34,19 +34,73 @@ export const CSVDownload: Dataset = {
...
+# Starlark Documentation: qri.io/docs
+# Transform Code Snippet Library: qri.io/docs
+`
+      },
+      {
+        syntax: 'starlark',
+        script: `# CSV Download Code Sample
+# This really works! Click 'Dry Run' to try it ↗
+
+# import Starlark dependencies
```
b5: can we change this to just "import dependencies"?

b5 (approved):
two nits, then looks good

PR #343, opened 2021-09-23
feat: various style tweaks to improve the workflow editor

chriswhong:
Contains lots of style tweaks to make the header, dataset menu, and workflow editor more closely match the mockups.

Fixes a resize bug with the code editors (they would grow, but not shrink when resizing the window.

image


b5 (approved):
rebased. LGTM!

PR #346, opened 2021-09-24
feat: updated workflow UI to match new design

boandriy:
updated workflow UI to match new design

image


src/features/trigger/CronTrigger.tsx
```diff
@@ -1,34 +1,49 @@
...
+
+  }
+
+  const handleTriggerChange = (t: WorkflowTrigger) => {
+    setTriggers([t])
   }
 
   // split the ISO8601 repeating interval (R/{starttime}/{interval})
-  const [, startTime, interval] = trigger.periodicity.split('/')
+  const [, startTime, interval] = triggers[0].periodicity.split('/')
```
ramfox: I'm getting typescript error on the `periodicity` field.

`triggers` is an array of `WorkflowTrigger`. The `WorkflowTrigger` interface does not have `periodicity` as a properly. We need to first cast `trigger[0]` as a `CronTrigger` before trying to get `periodicity` from it:

```
 const cronTrigger = triggers[0] as CronTrigger
 const [, startTime, interval] = cronTrigger.periodicity.split('/')
```

src/features/trigger/CronTrigger.tsx
```diff
@@ -50,25 +65,33 @@ const CronTrigger: React.FC<CronTriggerProps> = ({
...
-          >Edit</Link>
+          </div>}
         </div>
       </div>
-    </Block>
+      {isEditable &&
+      <div className='w-full'>
+        <div className='text-qrigray-700 text-xs'>Repeat</div>
+        <CronTriggerEditor
+          trigger={triggers[0]}
```
ramfox: getting a typescript error here as well. Once we cast `trigger[0]` as a `CronTrigger`, we can use that same variable to pass as the `trigger` into `CronTriggerEditor`.

src/features/trigger/WorkflowTriggersEditor.tsx
```diff
@@ -25,12 +26,22 @@ const WorkflowTriggersEditor: React.FC<WorkflowTriggersEditorProps> = ({
 }) => {
   const dispatch = useDispatch()
 
+  const [ showControls, setShowControls ] = useState(false);
+  const [ createTrigger, setCreateTrigger ] = useState(false)
+
+  const defaultCronTrigger: WorkflowTrigger = {
+    type: 'cron',
+    periodicity: 'R/2021-01-01T02:00:00.000Z/P1D'
+  }
```
ramfox: The type here should be `CronTrigger`

```
const defaultCronTrigger: CronTrigger = {
    type: 'cron',
    periodicity: 'R/2021-01-01T02:00:00.000Z/P1D'
  }
```

src/features/workflow/state/workflowState.ts
```diff
@@ -8,7 +8,10 @@ import {
   SetWorkflowRefAction,
   WorkflowTriggerAction,
   RunModeAction,
-  SetTemplateAction, AddWorkflowStepAction, RemoveWorkflowStepAction
+  SetTemplateAction,
+  AddWorkflowStepAction,
+  RemoveWorkflowStepAction,
```
ramfox: getting error from my linter

`Module '"./workflowActions"' has no exported member 'RemoveWorkflowStepAction'.ts(2305)`

ramfox (approved):
There are a few linting errors that I'm finding, once those are fixed & pushed this LGTM!

PR #347, opened 2021-09-27
fix: Display username and dataset name in toast as link to a ds preview

### LINKED ISSUE #261: Toast identifies the dataset that is running
We should be able to see the `username/name`, and perhaps the run number when a toast shows up.  On success or failure, clicking the toast (or a link in the toast) should go to the run log.

boandriy:
Display username and dataset name in toast as link to a ds preview
Edited the collection state to include runs version info object

image

Fixes #261

### DIFF
--- a/src/features/collection/state/collectionActions.ts
+++ b/src/features/collection/state/collectionActions.ts
@@ -3,6 +3,8 @@ import { newVersionInfo, VersionInfo } from "../../../qri/versionInfo"
 import { NewWorkflow, workflowInfoFromWorkflow } from "../../../qrimatic/workflow"
 import { WORKFLOW_COMPLETED, WORKFLOW_STARTED } from "./collectionState"
 import { AnyAction } from "redux"
+import { RootState } from "../../../store/store";
+import { ThunkDispatch } from 'redux-thunk'
 
 function mapVersionInfo (data: object | []): VersionInfo[] {
   if (!data) { return [] }
@@ -13,6 +15,18 @@ export function loadCollection() {
   return async (dispatch, getState) => {
     return dispatch(fetchCollection())
   }
+}
+
+export function loadVersionInfo(initID: string) {
+  return async (dispatch: ThunkDispatch<any, any, any>, getState: () => RootState) => {
+    const state = getState();
+    if (state.collection.collection[initID]) {
+      return state.collection.collection[initID]
+    } else if (!state.collection.pendingIDs.includes(initID)) {
+      return dispatch(fetchVersionInfo(initID))
+    } else return
+  }
+
 }
 
 function fetchCollection (): ApiAction {
@@ -26,6 +40,19 @@ function fetchCollection (): ApiAction {
         limit: 300 // TODO(chriswhong): For now, we assume the client has the entire collection and can sort and filter locally
       },
       map: mapVersionInfo
+    }
+  }
+}
+
+function fetchVersionInfo (initID: string): ApiAction {
+  return {
+    type: 'versioninfo',
+    [CALL_API]: {
+      endpoint: 'collection/get',
+      method: 'POST',
+      body: {
+        initID: initID
+      },
     }
   }
 }
--- a/src/features/collection/state/collectionState.ts
+++ b/src/features/collection/state/collectionState.ts
@@ -1,6 +1,7 @@
 import { RootState } from '../../../store/store';
 import { createReducer } from '@reduxjs/toolkit'
-import { datasetAliasFromVersionInfo, VersionInfo } from '../../../qri/versionInfo';
+import { datasetAliasFromVersionInfo, newVersionInfo, VersionInfo } from '../../../qri/versionInfo';
+import { CALL_API } from "../../../store/api";
 
 export const WORKFLOW_STARTED = 'WORKFLOW_STARTED'
 export const WORKFLOW_COMPLETED = 'WORKFLOW_COMPLETED'
@@ -18,7 +19,7 @@ export const selectCollection = (state: RootState): VersionInfo[] => {
     ordered.push(collection[id])
   })
 
-  Object.keys(collection).forEach( (id: string) => {
+  state.collection.listedIDs.forEach( (id: string) => {
     if (running.includes(id)) {
       return
     }
@@ -31,6 +32,9 @@ export const selectCollection = (state: RootState): VersionInfo[] => {
   })
 }
 
+export const selectVersionInfo = ( initId: string): (state: RootState) => VersionInfo =>
+  (state) => state.collection.collection[initId]
+
 export const selectIsCollectionLoading = (state: RootState): boolean => state.collection.collectionLoading && state.collection.runningLoading
 
 export interface CollectionState {
@@ -39,26 +43,36 @@ export interface CollectionState {
   // running contains the ids of the currently running workflows in reverse chronological
   // order based on latestRunTime
   running: string[]
+  // ids of datasets in the current user's collection
+  listedIDs: Set<string>
+  // ids of run events in middle of loading
+  pendingIDs: string[]
   collectionLoading: boolean
   runningLoading: boolean
+
 }
 
 const initialState: CollectionState = {
   collection: {},
   running: [],
+  listedIDs: new Set<string>(),
+  pendingIDs: [],
   collectionLoading: true,
   runningLoading: true
 }
 
 export const collectionReducer = createReducer(initialState, {
   'API_COLLECTION_REQUEST': (state, action) => {
-    state.collection = {}
+    state.listedIDs = new Set<string>()
     state.collectionLoading = true
   },
   'API_COLLECTION_SUCCESS': (state, action) => {
+    const listedIDs:Set<string> = new Set()
     action.payload.data.forEach((d: VersionInfo) => {
       state.collection[d.initID] = d
+      listedIDs.add(d.initID)
     })
+    state.listedIDs = listedIDs
     state.collectionLoading = false
   },
   'API_COLLECTION_FAILURE': (state, action) => {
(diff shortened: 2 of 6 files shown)


src/features/toast/Toast.tsx
```diff
@@ -23,6 +25,10 @@ const Toast: React.FC<ToastProps> = ({ message, type, id }) => {
           'text-qrigreen': type === 'succeeded',
           'text-qrired-700': type === 'failed',
         })}>{message}</div>
+        <Link to={'/ds/'+name}
+              className='text-qrinavy w-56 whitespace-nowrap overflow-hidden block text-sm overflow-ellipsis'>
+          {name}
+        </Link>
         <div className='text-xs'>{id}</div>
```
ramfox: we should remove the id now that we have the dataset reference hooked up correctly!

src/features/toast/Toast.tsx
```diff
@@ -2,15 +2,17 @@ import React from 'react'
 import classNames from 'classnames'
 import RunStatusIcon from '../run/RunStatusIcon'
 import IconButton from '../../chrome/IconButton'
+import { Link } from "react-router-dom";
 
 interface ToastProps {
   message: string
+  name: string
```
ramfox: `name` should be renamed to `ref`. 

In Qri, the `name` refers to just the name of the dataset. In this case, we are sending the `username/name` combination, which we call the "dataset reference", usually shortened to `ref`.

(Technically, the dataset reference can be as short as the `username/name` combo or as long as `username/name@initID/network/path`, etc, but is usually just the `username/name`)

src/features/toast/ToastRenderer.tsx
```diff
@@ -1,41 +1,85 @@
...
+        const initialResponse = await fetch(`${API_BASE_URL}/collection/get`, {
+          method: 'POST',
+          body: JSON.stringify({initID: initId})
+        })
+        const response:VersionInfo = await initialResponse.json().then(r => r.data)
+        if(response){
+          setCollections([...collections, response])
+        }
+      }
+    })
```
ramfox: usually when we see fetching happening in the component along the use of `useState`, it usually means we are missing something in redux. I proposed a large-ish refactor here that changes the way we think of the `state.collection.collection` & uses it as a place to store version infos, rather than the place that tells us what collections to show on the "My Collections" page. It would also mean a bunch of changes to this `ToastRenderer` component: https://github.com/qri-io/frontend/issues/261#issuecomment-928447901

LMK if you have any questions!

ramfox (changes requested):
I think we are missing some things in our current redux store that is making the patterns we have in the `ToastRenderer` component not as clean as we would like them to be. Take a look at my comment here https://github.com/qri-io/frontend/pull/347#discussion_r717117719 and let me know if you have any questions

src/features/collection/state/collectionState.ts
```diff
@@ -31,6 +32,9 @@ export const selectCollection = (state: RootState): VersionInfo[] => {
   })
 }
 
+export const selectVersionInfo = ( initId: string): (state: RootState) => VersionInfo =>
```
ramfox, line 35: `( initId: string)` => `(initId: string)`

src/features/collection/state/collectionState.ts
```diff
@@ -84,6 +94,14 @@ export const collectionReducer = createReducer(initialState, {
...
     state.runningLoading = false
   },
+  'API_VERSIONINFO_REQUEST' : (state, action) => {
+    state.pendingIDs.push(action[CALL_API].body.initID)
+  },
+  'API_VERSIONINFO_SUCCESS':(state, action) => {
+    const versionInfo: VersionInfo = action.payload.data;
+    state.pendingIDs = state.pendingIDs.filter(id => id !== versionInfo.initID)
+    state.collection[versionInfo.initID] = newVersionInfo(versionInfo)
+  },
```
ramfox, line 108: should add a failure case here to remove the initID from the `pending` list. If this causes an infinite loop we should talk over how to fix it.

src/features/toast/Toast.tsx
```diff
@@ -2,15 +2,27 @@ import React from 'react'
...
+import { VersionInfo } from "../../qri/versionInfo";
 
 interface ToastProps {
   message: string
+  initID:string
   type: 'running' | 'succeeded' | 'failed'
-  id: string
 }
 
+const getRef = (vi: VersionInfo | undefined) => {
```
ramfox: nice, we already have something like this:
https://github.com/qri-io/frontend/blob/6726bc5c4f653ee66eb7dd3f362fa640e46115c3/src/qri/versionInfo.ts#L94-L96

We should modify it so that if there is no username or name then we return an empty string (rather than '/')

src/features/toast/ToastRenderer.tsx
```diff
@@ -1,41 +1,49 @@
...
+  const initIDs = runs.map((r: Run) => r.initID)
 
   // artifically delay the update to 'success'
   const SUCCESS_START_DELAY = 0
 
   // after 'success', close the toast 4 seconds later
   const SUCCESS_CLOSE_DELAY = 4000
 
+  useEffect(() => {
+    initIDs.forEach((initId) => {
```
ramfox: we should be adding type to as much as we can:

`initIDs.forEach((initID: string) => {`

src/features/collection/state/collectionActions.ts
```diff
@@ -15,6 +16,18 @@ export function loadCollection() {
...
 }
 
+export function loadVersionInfo(initID: string) {
+  return async (dispatch, getState: () => RootState) => {
+    const state = getState();
+    if(state.collection.collection[initID]){
+      return state.collection.collection[initID]
+    }else if(!state.collection.pendingIDs.includes(initID)){
+      return dispatch(fetchVersionInfo(initID))
+    }else return
```
ramfox: some linting errors here:

```
    if (state.collection.collection[initID]) {
      return state.collection.collection[initID]
    } else if (!state.collection.pendingIDs.includes(initID)) {
      return dispatch(fetchVersionInfo(initID))
    } else return
```

ramfox (changes requested):
Hey hey!

Looks great!!!

I have a view renaming/linting comments.

I'm getting these `eslint` warning when I run `yarn start`

```
src/features/app/modal/Modal.tsx
  Line 1:49:  'useState' is defined but never used  @typescript-eslint/no-unused-vars

src/features/commits/CommitSummaryHeader.tsx
  Line 14:19:  'path' is assigned a value but never used  @typescript-eslint/no-unused-vars

src/features/dataset/DatasetRoutes.tsx
  Line 35:5:  React Hook useEffect has a missing dependency: 'qriRef'. Either include it or remove the dependency array  react-hooks/exhaustive-deps

src/features/dsPreview/state/dsPreviewActions.ts
  Line 1:18:  'refStringFromQriRef' is defined but never used  @typescript-eslint/no-unused-vars

Search for the keywords to learn more about each warning.
To ignore, add // eslint-disable-next-line to the line before.
```

I think these may have been fixed in previous prs tho, maybe they will go away after rebasing?

src/features/collection/state/collectionState.ts
boandriy: @ramfox , it causes infinite loop.

ramfox: okay fab. After looking into this more, I think we have some bigger structural issues to deal with that will help fix this problem, so we should just leave it alone for now and move on.

src/features/collection/state/collectionActions.ts
b5: if statements are still missing spaces.

src/features/collection/state/collectionState.ts
b5: still not resolved. check spacing around leading parenthesis

src/features/collection/state/collectionState.ts
```diff
@@ -39,25 +43,31 @@ export interface CollectionState {
   // running contains the ids of the currently running workflows in reverse chronological
   // order based on latestRunTime
   running: string[]
+  listedIDs: string[]
+  pendingIDs: string[]
```
b5: I'd appreciate comments on what these two lists are doing, similar to the running field above:

```ts
// ids of datasets in the current user's collection
listedIDs: Set<string>
// ids of datasets not (or not yet) in the user's collection
pendingIDs: string[]
```

src/features/collection/state/collectionState.ts
```diff
@@ -39,25 +43,31 @@ export interface CollectionState {
...
   'API_COLLECTION_REQUEST': (state, action) => {
-    state.collection = {}
+    state.listedIDs = []
     state.collectionLoading = true
   },
   'API_COLLECTION_SUCCESS': (state, action) => {
     action.payload.data.forEach((d: VersionInfo) => {
       state.collection[d.initID] = d
+      state.listedIDs.push(d.initID)
     })
```
b5: this leaves open the chance of repeating IDs in the `listedIDs` array. What we need is a set union between current and new values, and the easiest way to do that is to make the `listedIDs` field a `Set<string>` type. You'll need to use `.add` instead of `.push`

```ts
 action.payload.data.forEach((d: VersionInfo) => {
      state.collection[d.initID] = d
      state.listedIDs.add(d.initID)
  })
```

src/features/collection/state/collectionState.ts
```diff
@@ -39,25 +43,31 @@ export interface CollectionState {
...
+  pendingIDs: string[]
   collectionLoading: boolean
   runningLoading: boolean
+
 }
 
 const initialState: CollectionState = {
   collection: {},
   running: [],
+  listedIDs: [],
```
b5:
```suggestion
  listedIDs: new Set(),
```

src/features/collection/state/collectionState.ts
```diff
@@ -39,25 +43,31 @@ export interface CollectionState {
...
+  pendingIDs: [],
   collectionLoading: true,
   runningLoading: true
 }
 
 export const collectionReducer = createReducer(initialState, {
   'API_COLLECTION_REQUEST': (state, action) => {
-    state.collection = {}
+    state.listedIDs = []
```
b5: `state.listedIDs = new Set()`

b5 (changes requested):
coming along nicely! A few things to clean up

src/features/collection/state/collectionActions.ts
b5: _still_ missing. Maybe the wrong rule is writing on save in your IDE?

boandriy: @b5 ,  sorry.. fixed ^

ramfox (approved):
LGTM! Great work!

PR #352, opened 2021-09-29
fix: after navigating from new ds workflow to existing ds workflow, code no longer persists

### LINKED ISSUE #326: Workflow Steps Code persists when navigating from in-progress new workflow to existing dataset workflow
Observed just now in rosebud:

- Use "New Dataset", edit a cell
- Navigate to collection (accepting unsaved changes warning)
- Select a dataset with an existing workflow, navigate to its workflow route
- Cells reflect the code from the New Dataset

boandriy:
fixed issue where after changing code in new ds workflow you would navigate to existing ds workflow and cells would not change

Fixes #326

### DIFF
--- a/src/features/workflow/state/workflowState.ts
+++ b/src/features/workflow/state/workflowState.ts
@@ -10,7 +10,8 @@ import {
   RunModeAction,
   SetTemplateAction,
   AddWorkflowStepAction,
-  WorkflowStepAction
+  WorkflowStepAction,
+  UndoWorkflowChanges
 } from './workflowActions'
 import { NewRunFromEventLog, Run } from '../../../qri/run'
 import { Workflow, WorkflowBase } from '../../../qrimatic/workflow'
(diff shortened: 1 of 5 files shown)


src/features/workflow/state/workflowState.ts
```diff
@@ -289,6 +292,17 @@ function moveWorkflowTransformStepUp (state: WorkflowState, action: WorkflowStep
   return
 }
 
+function workflowUndoChanges (state: WorkflowState, actions: UndoWorkflowChanges) {
+  const oldTriggers:WorkflowTrigger[] | undefined = JSON.parse(JSON.stringify(state.workflowBase.triggers))
+  const oldHooks: WorkflowHook[] | undefined = JSON.parse(JSON.stringify(state.workflowBase.hooks))
+  const oldSteps: TransformStep[] = JSON.parse(JSON.stringify(state.workflowBase.steps || []))
```
b5: why do we need this double JSON parse/unparse?

src/features/workflow/state/workflowState.ts
```diff
@@ -289,6 +292,17 @@ function moveWorkflowTransformStepUp (state: WorkflowState, action: WorkflowStep
...
+  state.workflow.triggers = oldTriggers
+  state.workflow.hooks = oldHooks
+  if(state.dataset.transform){
```
b5: if statement spacing

src/features/workflow/state/workflowState.ts
boandriy: its a trick to make a deep copy of an object

ramfox (approved):
LGTM!

Before you merge, let's change the title of this PR to be more about the _effect_ of the change you are making, rather than _why_ you are making it. To say it another way, when we make bug fixes, the commits and PR titles we write should be about "what happens now" rather than "what was wrong before".

example:
`fix(workflow): do not keep code changes after clicking "discard changes"`

PR #356, opened 2021-09-30
fix: ds name can be edited when you go to workflow route

### LINKED ISSUE #293: name editing no longer works
I looks like we have lost the ability to click on a dataset name in the header and change it.  Investigate and restore.

boandriy:
fixed issue where ds name was not editable

fixes #293


ramfox (approved):
LGTM!

PR #357, opened 2021-09-30
feat: added a delete confirmation modal for cell delete

### LINKED ISSUE #344: Confirm cell deletion
If a user clicks "delete" in a workflow cell's contextual menu, it is instantly deleted.  We need to add a confirmation step or an undo (cmd-Z).  Deepnote, our inspiration for much of this functionality, does not have a confirmation step, but does have undo.

boandriy:
added a delete confirmation modal for cell delete

image

fixes #344


chriswhong:
Let's hold on this until we have a mockup, I think a modification of the delete button to prompt the user may be a better approach than a modal.

PR #358, opened 2021-10-01
feat: added cell status UI in workflow outline

### LINKED ISSUE #349: Show code block glyphs in outline
Each code block should be represented by a glyph/row in the outline, where we can also show the block's status when Dry running the script.

We lost this functionality when we switched to free-form code cells because the cells no longer have names.  We can bring them back without names, the user will just be able to tell at a glance how many blocks there are and visualize the dry-run as it makes its way through the sequence

boandriy:
added cell status UI in workflow outline

image

closes #349


ramfox (approved):
LGTM!

Issue #313, opened 2021-09-08
Workflow Cell output is cleared after editing code


chriswhong:
This is likely being cleared when we calculate `isDirty` for the workflow, but we should leave it around as it is helpful in debugging.

My story:  I am trying to transform several columns in each row of the dataset, and am using `print()` to log my work.  I need to identify the columns so I can pull items out of a list, but when I start writing code, the output disappears and I need to dry run again to find the next column and continue writing code.

boandriy:
Hey, @b5 , @ramfox , the reason why the output is cleared is that on cell script edit we call `changeWorkflowTransformStep`: 
```typescript
function changeWorkflowTransformStep(state: WorkflowState, action: SetWorkflowStepAction) {
  if (state.dataset.transform?.steps) {
    state.dataset.transform.steps[action.index].script = action.script
  }

  state.isDirty = calculateIsDirty(state)
  // clear out events after an edit to reset dry run RunStatus
  state.events = []

  return
}
```

Here we are clearing `state.events` - which leads to clear of cells outputs too.

The simplest way of solving this issue is just removing the line of code that clears the events state. But it will not only leave the output UI unchanged on edit, it will also leave the status of each step unchanged and dry run status unchanged.

The other way to solve this is - adding workflow state with output variable that would be an array of outputs. And using this new state variable to display the output to cells.

ramfox:
@boandriy let's talk about this when I'm in on thursday, it relates to the issue I just filed about storing events in a different place in the state tree now that we are expecting automated runs to happen while we are editing a different workflow: #350 

But I think your instinct is right: we need a different way of storing and referring to events. We need to be able to keep the previous run & the current/next run.

But it would also be great to get some clarity around what the product team feels is the correct UI interaction.

@chriswhong, when you start editing the cell, do you expect all the previous run information to stay until you click "deploy" or "dry run" again (not only the cell output, but also the sidebar status and coloring around the cell), or do you expect the previous run information (everything besides the cell output) to be cleared when you go to edit the cell?

chriswhong:
There has been some debate on this.  I think once you start editing, the success/failure state of a given block should be thrown out, and the rest should be retained until the next Dry Run.  This is what deepnote does and is worth emulating.

chriswhong:
I think in addition to this, the "overall" state of the script should also reset, or maybe show some kind of "dirty" icon, as it's success/failure is also no longer valid once a cell has been edited.

ramfox:
[...]

PR #360, opened 2021-10-04
refactor: moved events to its own state tree section from workflow state tree

### LINKED ISSUE #350: refactor(state): events need their own section of the state tree
Right now, all events are being stored in the `workflow` section of the state tree. While we definitely need the run info for the current workflow in `workflow` section of the state tree, but we are also mixing that with events that come from other datasets' runs.

We need an `events` section of the state tree & selectors that allow you to select events based on `sessionID`. We can use that `sessionID` to select events based on `runID`, `ref`, `initID` etc. We can also add selectors as needed to filter events further.

The `workflow` reducer can react to events by selecting them from the `events` section of the state tree.

boandriy:
moved events to its own state tree section from workflow as not all events are tied to the workflow. 

closes #350

### DIFF
--- a/src/features/activityFeed/DatasetActivityFeed.tsx
+++ b/src/features/activityFeed/DatasetActivityFeed.tsx
@@ -8,8 +8,9 @@ import { loadDatasetLogs } from './state/activityFeedActions'
 import { newDatasetLogsSelector } from './state/activityFeedState'
 import DatasetFixedLayout from '../dataset/DatasetFixedLayout'
 import { runNow } from '../workflow/state/workflowActions'
-import { selectLatestRun } from '../workflow/state/workflowState'
+import { selectLatestRunId } from '../workflow/state/workflowState'
 import RunNowButton from './RunNowButton'
+import { selectRun } from "../events/state/eventsState";
 
 
 export interface DatasetActivityFeedProps {
@@ -21,7 +22,8 @@ const DatasetActivityFeed: React.FC<DatasetActivityFeedProps> = ({
 }) => {
 
   const logs = useSelector(newDatasetLogsSelector(qriRef))
-  const latestRun = useSelector(selectLatestRun)
+  const latestRunId = useSelector(selectLatestRunId)
+  const latestRun = useSelector(selectRun(latestRunId))
   const dispatch = useDispatch()
 
   const [tableContainer, { height: tableContainerHeight }] = useDimensions()
--- a/src/features/events/state/eventsActions.ts
+++ b/src/features/events/state/eventsActions.ts
@@ -1,0 +1,31 @@
+import { EventLogLine } from "../../../qri/eventLog";
+
+import {
+  REMOVE_EVENT,
+  RUN_EVENT_LOG
+} from "./eventsState";
+
+export interface EventLogAction {
+  type: string
+  data: EventLogLine
+}
+
+export interface RemoveEventAction {
+  type: string
+  sessionID: string
+}
+
+export function runEventLog (event: EventLogLine): EventLogAction {
+  return {
+    type: RUN_EVENT_LOG,
+    data: event
+  }
+}
+
+export function removeEvent(sessionID: string): RemoveEventAction {
+  return {
+    type: REMOVE_EVENT,
+    sessionID: sessionID
+  }
+}
+
--- a/src/features/events/state/eventsState.ts
+++ b/src/features/events/state/eventsState.ts
@@ -1,0 +1,44 @@
+import { createReducer } from '@reduxjs/toolkit'
+
+import { EventLogLine } from "../../../qri/eventLog";
+import { NewRunFromEventLog, Run } from "../../../qri/run";
+import { RootState } from "../../../store/store";
+import { EventLogAction, RemoveEventAction } from "./eventsActions";
+
+export const RUN_EVENT_LOG = 'RUN_EVENT_LOG'
+export const REMOVE_EVENT = 'REMOVE_EVENT'
+
+export const selectRuns = (state: RootState): Run[] => {
+  const { events } = state.events
+  const unique = Array.from(new Set(events.map(d => d.sessionID)))
+  return unique.map((d) => {
+    return NewRunFromEventLog(d, events.filter(e => e.data.mode !== 'apply'))
+  })
+}
+
+export const selectRun = (sessionId: string): (state: RootState) =>  Run =>
+  (state:RootState) => NewRunFromEventLog(sessionId, state.events.events)
+
+export interface EventsState {
+  events: EventLogLine[]
+}
+
+const initialState: EventsState = {
+  events: [],
+}
+
+export const eventsReducer = createReducer(initialState, {
+  RUN_EVENT_LOG: addRunEvent,
+  REMOVE_EVENT: removeEvent,
+})
+
+
+function addRunEvent(state: EventsState, action: EventLogAction) {
+  state.events.push(action.data)
+  state.events.sort((a,b) => a.ts - b.ts)
+}
+
+function removeEvent(state: EventsState, action: RemoveEventAction) {
+  state.events = state.events.filter(e => e.sessionID !== action.sessionID)
+}
+
--- a/src/features/workflow/CodeEditor.tsx
+++ b/src/features/workflow/CodeEditor.tsx
@@ -81,8 +81,8 @@ const CodeEditor: React.FC<CodeEditorProps> = ({
     }
   }, [theEditor])
 
-  const handleEditorDidMount = (editor: MonacoEditor['editor']) => {
-    editor?.addAction({
+  useEffect(() => {
+    theEditor?.addAction({
       id: "executeCurrentAndAdvance",
       label: "Execute Block and Advance",
       keybindings: [KeyMod.CtrlCmd | KeyCode.Enter],
@@ -90,9 +90,11 @@ const CodeEditor: React.FC<CodeEditorProps> = ({
       precondition: "editorTextFocus && !suggestWidgetVisible && !renameInputVisible && !inSnippetMode && !quickFixWidgetVisible",
       run: () => {
         onRun()
-      },
-    });
+      }
+    })
+  }, [theEditor, onRun])
 
+  const handleEditorDidMount = (editor: MonacoEditor['editor']) => {
     setTheEditor(editor)
   }
 
--- a/src/features/workflow/RunBar.tsx
+++ b/src/features/workflow/RunBar.tsx
@@ -12,13 +12,16 @@ import {
   selectWorkflow,
   selectWorkflowDataset,
   selectApplyStatus,
-  selectWorkflowIsDirty
+  selectWorkflowIsDirty,
+  selectLatestDryRunId,
+  selectLatestRunId
 } from './state/workflowState'
 import { platform } from '../../utils/platform'
 import Icon from "../../chrome/Icon";
 import DeployButton from "../deploy/DeployButton";
 import { newQriRef } from "../../qri/ref";
 import { useParams } from "react-router";
+import { removeEvent } from "../events/state/eventsActions";
 
 export interface RunBarProps {
  status: RunStatus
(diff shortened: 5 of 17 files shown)


src/features/activityFeed/DatasetActivityFeed.tsx
```diff
@@ -21,7 +22,7 @@ const DatasetActivityFeed: React.FC<DatasetActivityFeedProps> = ({
 }) => {
 
   const logs = useSelector(newDatasetLogsSelector(qriRef))
-  const latestRun = useSelector(selectLatestRun)
+  const latestRun = useSelector(selectRun(useSelector(selectLatestRunId)))
```
ramfox, line 25: For clarify, probably best to switch this to two lines:

```
const latestRunID = useSelector(selectLatestRunId))
const latestRun = useSelector(selectorRun(latestRunID))
```

ramfox: For clarify, probably best to switch this to two lines:

```
const latestRunID = useSelector(selectLatestRunId))
const latestRun = useSelector(selectorRun(latestRunID))
```

src/features/events/state/eventsActions.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export function runEventLog (event: EventLogLine): EventLogAction {
+  return {
+    type: RUN_EVENT_LOG,
+    data: event
+  }
+}
+
+export function removeEvent(sessionID: string):RemoveEventAction {
```
ramfox: missing a space between the colon and `RemoveEventAction`
`(sessionID: string):RemoveEventAction` => `(sessionID: string): RemoveEventAction`

src/features/events/state/eventsState.ts
```diff
@@ -0,0 +1,44 @@
...
+export const selectRuns = (state: RootState): Run[] => {
+  const { events } = state.events
+  const unique = Array.from(new Set(events.map(d => d.sessionID)))
+  return unique.map((d) => {
+    return NewRunFromEventLog(d, events.filter(e => e.data.mode !== 'apply'))
+  })
+}
+
+export const selectRun = (sessionId: string | undefined): (state: RootState) =>  Run | undefined =>
+  (state:RootState) => sessionId ? NewRunFromEventLog(sessionId, state.events.events) : undefined
```
ramfox: let's refactor this so that it accepts only a `string` (not a `string | undefined`) and adjust `selectLatestRunID` so that it returns an empty string if there is no latest run (rather than `undefined`)

src/features/workflow/RunBar.tsx
```diff
@@ -35,9 +38,19 @@ const RunBar: React.FC<RunBarProps> = ({
...
   const applyStatus = useSelector(selectApplyStatus)
   const isDirty = useSelector(selectWorkflowIsDirty)
+  const latestDryRunId = useSelector(selectLatestDryRunId)
+  const latestRunId = useSelector(selectLatestRunId)
   const qriRef = newQriRef(useParams())
 
+  const removeRunEvents = () => {
+    if (latestDryRunId)
+      dispatch(removeEvent(latestDryRunId))
+    if(latestRunId)
```
ramfox: `if(latestRunId)` => `if (latestRunId)`

src/features/workflow/Workflow.tsx
```diff
@@ -34,7 +35,7 @@ const Workflow: React.FC<WorkflowProps> = ({ qriRef }) => {
   const location = useLocation<WorkflowLocationState>()
   const workflow = useSelector(selectWorkflow)
   const workflowDataset = useSelector(selectWorkflowDataset)
-  const latestRun = useSelector(selectLatestDryRun)
+  const latestRun = useSelector(selectRun(useSelector(selectLatestDryRunId)))
```
ramfox, line 38: let's split into two calls:

```
const latestRunID = useSelector(selectLatestDryRunId))
const latestRun = useSelector(selectRun(latestRunID))
```

src/features/workflow/WorkflowCellControls.tsx
```diff
@@ -2,23 +2,25 @@ import React from "react";
...
   removeWorkflowTransformStep
 } from './state/workflowActions';
 import WorkflowCellControlButton from "./WorkflowCellControlButton";
+import { removeEvent } from "../events/state/eventsActions";
 
 
 interface WorkflowCellControlsProps {
   index: number
   setAnimatedCell: (i:number) => void
+  sessionId: string | undefined
```
ramfox: let's force this to only accept a `string`, rather than `string | undefined` & make whatever changes to the call site so that we are always passing in a string.

src/features/workflow/WorkflowCellControls.tsx
```diff
@@ -32,12 +34,17 @@ const WorkflowCellControls: React.FC<WorkflowCellControlsProps> = ({
     dispatch(duplicateWorkflowTransformStep(index))
   }
 
+  const onOutputClear = () => {
+    if(sessionId)
```
ramfox: `if(sessionId)` => `if (sessionId)`

src/features/workflow/WorkflowEditor.tsx
```diff
@@ -145,7 +148,11 @@ const WorkflowEditor: React.FC<WorkflowEditorProps> = ({
                     onChangeScript={(i:number, script:string) => {
                       if (dataset?.transform?.steps) {
                         dispatch(changeWorkflowTransformStep(i, script))
+                        if (run && run.id) {
+                          dispatch(removeEvent(run.id))
+                        }
```
ramfox: If we clear the run here, then we won't have the ability to display the output.

We should remove these lines, and instead call `removeEvent` in `runScript`, before we actually dispatch `applyWorkflowTransform`

We also need to make sure we `removeEvent` any place that we "dry run" or "deploy"!

src/features/workflow/WorkflowPage.tsx
```diff
@@ -19,7 +20,7 @@ interface WorkflowPageProps {
 const WorkflowPage: React.FC<WorkflowPageProps> = ({ qriRef }) => {
   const dispatch = useDispatch()
   let dataset = useSelector(selectWorkflowDataset)
-  const latestRun = useSelector(selectLatestDryRun)
+  const latestRun = useSelector(selectRun(useSelector(selectLatestDryRunId)))
```
ramfox: separate into two lines, one selecting `latestRunID` and the other selecting `latestRun`

src/features/workflow/state/workflowState.ts
```diff
@@ -39,28 +34,20 @@ export const SET_RUN_MODE = 'SET_RUN_MODE'
...
 }
 
-export const selectLatestRun = (state: RootState): Run | undefined => {
+export const selectLatestRunId = (state: RootState): string | undefined => {
   if (state.workflow.lastRunID) {
-    return NewRunFromEventLog(state.workflow.lastRunID, state.workflow.events)
+    return state.workflow.lastRunID
   }
   return undefined
 }
```
ramfox: let's refactor these to return a `string` only. If there is no `lastDryRunID` or `lastRunID`, return the empty string.

ramfox (changes requested):
Looking good!

I have some requested changes, but overall nice work.

Most changes are linting changes/minor refactors. One bigger change is in `WorkflowEditor`. Right now, we are removing the run each time the script changes, we should instead be removing the run when we go to dry run or run the script again.

src/features/workflow/CodeEditor.tsx
```diff
@@ -81,6 +81,19 @@ const CodeEditor: React.FC<CodeEditorProps> = ({
...
+      label: "Execute Block and Advance",
+      keybindings: [KeyMod.CtrlCmd | KeyCode.Enter],
+      contextMenuGroupId: "2_execution",
+      precondition: "editorTextFocus && !suggestWidgetVisible && !renameInputVisible && !inSnippetMode && !quickFixWidgetVisible",
+      run: () => {
+        onRun()
+      }
+    })
+  }, [theEditor, onRun])
+
```
ramfox: this is new functionality, correct?

If so, add a comment in the main PR comment saying that you are adding the ability to dry run a workflow by doing "cmd+enter" or "ctrl-enter" as part of this PR.

src/features/workflow/WorkflowEditor.tsx
```diff
@@ -71,6 +73,9 @@ const WorkflowEditor: React.FC<WorkflowEditorProps> = ({
   }, [dataset])
 
   const runScript = () => {
+    if (latestDryRunId) {
+      dispatch(removeEvent(latestDryRunId))
+    }
```
ramfox, line 79: This should also be cleared if there is a `latestRunId` as well.

This may be for the future, but we should look into if we can combine `latestRunID` and `latestDryRunID`, since we can only have one run at a time, I wonder if there is anything that relies on there being two different run ids.

ramfox (changes requested):
looking good!

We also need to clear events before we `deploy` a run. In the `DeployWorkflowModal`, we need to select the `latestDryRunId` and `latestRunId` & call `removeEvents` there before we deploy the workflow.

Edit: also it looks like this comment got skipped: https://github.com/qri-io/frontend/pull/360#discussion_r721509814

src/features/workflow/CodeEditor.tsx
boandriy: Its not a new functionality. It was there, but problem was that it was doing it on component mount. So latestDryRunId was neglected there. That is why I have added this code. @ramfox

src/features/workflow/WorkflowCellControls.tsx
```diff
@@ -32,12 +34,17 @@ const WorkflowCellControls: React.FC<WorkflowCellControlsProps> = ({
...
+      dispatch(removeEvent(sessionId))
+  }
+
   return (
     <div className={'group-hover:opacity-100 flex-grow-0 w-48 ml-8 opacity-0 transition-opacity relative'}>
       <div className={'flex flex-col bg-white w-48 rounded-md absolute p-5 pt-2.5'}>
         <WorkflowCellControlButton onClick={onDelete} label='Delete' icon='trashBin'/>
         <WorkflowCellControlButton onClick={onDuplicate} label='Duplicate block' icon='duplicate'/>
-        <WorkflowCellControlButton onClick={() => { dispatch(clearOutputWorkflowTransformStep(index)) }} label='Clean output' icon='circleX'/>
+        <WorkflowCellControlButton onClick={onOutputClear} label='Clean output' icon='circleX'/>
```
ramfox, line 47: ahhh, okay, i didn't understand what this was when I was reviewing before.

Based on the UI it looks like it's asking to "clean" the output of *just that cell*, while the way it is coded up is cleaning the output of every cell. 

I think we should adjust this for now. In the above, "outputClear" function, instead of clearing the events, just `console.log` saying "not yet implemented", and we should add this to the work you are doing on #313

boandriy: Yep, Will be done as part of that PR.

src/features/workflow/WorkflowEditor.tsx
ramfox: This looks like it got skipped!

src/features/workflow/CodeEditor.tsx
ramfox: ahh. Okay, so because `onRun` now changes when the `latestDryRunId` changes, `handleEditorDidMount` is not properly adding the `onRun` action because it only happens on mount.

Do we still need `handleEditorDidMount`? If so, can you add a comment to it explaining why we need both. I can see it being confusing in the future to see the same thing happening twice.

Also, why don't we need to call `setTheEditor` here?

src/features/workflow/WorkflowCellControls.tsx
ramfox: okay, as long as we know it's going to be fixed soon, this can stay as is.

src/features/workflow/CodeEditor.tsx
boandriy: @ramfox , did slight modifications here. We do still need the handleEditorDidMount as this is the only place we are initializing `theEditor` variable. If I call the `setTheEditor` here it breaks the UI changes by previous useEffect that determine the height of the cell. I think it is because we are not setting theEditor at any point of the component and working with the object itself. As I didn't really worked with monaco editor before I just thought that this is the way to work with it: just initialize it and then use internal method to make changes. @ramfox

ramfox (approved):
LGTM! great work!

PR #361, opened 2021-10-04
fix: ds validation error typo fixed

### LINKED ISSUE #354: dataset name validation error
Observed during a usability test:  User typed a datasetname that started with “D” but the error said “dataset names must start with a letter”

boandriy:
ds validation error typo fixed

image

fixes #354


boandriy:
@chriswhong , tagging only you for review here, as it is only a text change.

chriswhong (approved):
LGTM

PR #371, opened 2021-10-06
fix: added logic to hide the output and status of each cell/outline on edit while remaining the events list intact

### LINKED ISSUE #313: Workflow Cell output is cleared after editing code

boandriy:
added logic to hide the output and status of each cell/outline on edit while remaining the events list intact
fixed the output clear action to hide the output of single cell

fixes #313

### DIFF
--- a/src/features/workflow/CodeEditor.tsx
+++ b/src/features/workflow/CodeEditor.tsx
@@ -10,7 +10,7 @@ export interface CodeEditorProps {
   // determines whether the Component should render with rounded corners on the bottom
   standalone?: boolean
   onChange: (newValue: string) => void
-  hasOutput: boolean;
+  isEdited: boolean
   onRun?: () => void
   status?: RunStatus
 }
@@ -42,8 +42,8 @@ const CodeEditor: React.FC<CodeEditorProps> = ({
   onChange,
   onRun = () => {},
   disabled,
-  hasOutput,
-  status
+  status,
+  isEdited
 }) => {
   const ref = useRef<MonacoEditor>(null)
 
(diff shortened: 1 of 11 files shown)


src/features/workflow/CodeEditor.tsx
```diff
@@ -106,17 +108,20 @@ const CodeEditor: React.FC<CodeEditorProps> = ({
...
-    borderStyles = 'border-transparent hover:border-qritile border-2 border-solid'
+  let borderStyles = 'border-transparent hover:border-qritile border-2 border-solid';
+  if(!isEdited){
+    if(status === 'succeeded') {
+      borderStyles = `border-qrigreen border-2 border-solid ${hasOutput && 'border-b-0'}`
+    }else if(status === 'running'){
+      borderStyles = 'border-qrinavy-700 border-2 border-solid border-b-0 -mb-1'
+    }else if(status === 'failed'){
+      borderStyles = `border-dangerred border-2 border-solid ${hasOutput && 'border-b-0'}`
+    }else{
```
ramfox: Based on chris' answer on the issue I think the logic here isn't right (here is the comment: https://github.com/qri-io/frontend/issues/313#issuecomment-929369275)

When any cell as been edited, the *sidebar* and *all of the cells* should not have a "success" border any more & show a symbol that indicates that it is not in a valid state. In this comment, he mentions using the gray circle with the line as the symbol: https://github.com/qri-io/frontend/issues/313#issuecomment-929380395

ramfox (changes requested):
everything looks good, but I think the logic for the cell outlines is off, based on the comment's chris left on the issue

chriswhong:
image

When testing this locally, I see this green outline appearing on hover.  If this is the "active" state of the cell, it should not be related to hover and should just appear this way when the user has clicked it it and is ready to edit, regardless of whether they have made changes.

 On this note, we need a way for the current cell to stay active AND to show that its last run was successful.

On deepnote, they use the icon to indicate success, but there is no outline color for success/failure.  The outline is only used for the "active" cell that the user is interacting with.

When the user edits the cell, the results stay, but the icon disappears.

b5:
where are we at with this? need re-review? @chriswhong have you confirmed it works as expected?

chriswhong:
Yes, I pushed changes that are working as expected, just need the ok on code quality/maintainability

src/features/workflow/state/workflowState.ts
```diff
@@ -76,6 +85,12 @@ export interface WorkflowState {
   // stores the "clean" state of triggers, steps, and hooks, used to compare
   // with working state to determine isDirty
   workflowBase: WorkflowBase
+  // stores a boolean value for each workflow code cell to represent if
+  // cell was edited
+  areCellsEdited: boolean[]
```
b5: don't use `are` in a field name: `areCellsEdited` -> `editedCells`

src/features/workflow/state/workflowState.ts
```diff
@@ -76,6 +85,12 @@ export interface WorkflowState {
...
+  // stores a boolean value for each workflow code cell to represent if
+  // cell output was cleared
+  areCellsOutputCleared: boolean[]
```
b5: `areCellsOutpuCleared` -> `clearedOutputCells`

src/features/workflow/WorkflowCell.tsx
```diff
@@ -1,13 +1,17 @@
 import React from 'react'
+import classNames from 'classnames'
 import { RunStep } from '../../qri/run'
 import { TransformStep } from '../../qri/dataset'
 import CodeEditor from './CodeEditor'
 import Output from './output/Output'
 import ScrollAnchor from '../scroller/ScrollAnchor'
 import WorkflowHeader from './WorkflowHeader'
 import WorkflowCellControls from './WorkflowCellControls'
+import { useSelector } from "react-redux";
```
b5: keep npm / external package imports grouped at the top

PR #372, opened 2021-10-06
fix: fixed UI of deploy modal to fit all content

### LINKED ISSUE #370: Long content in deploy modal is hidden
The deploy modal is fixed height because it contains several screens in the same modal.  The warning gets bumped due to the content above.

I think we probably need to keep it fixed height but control the heights of the rest of the content to make sure there's always an empty spot for warnings.

image

boandriy:
fixed UI of deploy modal to fit all content
 
 fixes #370

### DIFF
--- a/src/features/workflow/modal/DeployModal.tsx
+++ b/src/features/workflow/modal/DeployModal.tsx
@@ -179,7 +179,7 @@ const DeployModal: React.FC = () => {
   }
 
   return (
-    <div style={{ width: 440, height: 440 }}>
+    <div style={{ width: 440, height: 510 }}>
       <div className={classNames('absolute w-full h-full bg-white p-8 transition-all duration-300 bg-white p-8 text-left text-black flex flex-col z-10', {
         'left-0': !deploying,
         '-left-full': deploying


src/features/workflow/modal/DeployModal.tsx
```diff
@@ -178,12 +179,13 @@ const DeployModal: React.FC = () => {
...
+  }
+
   return (
-    <div style={{ width: 440, height: 440 }}>
-      <div className={classNames('absolute w-full h-full bg-white p-8 transition-all duration-300 bg-white p-8 text-left text-black flex flex-col z-10', {
-        'left-0': !deploying,
-        '-left-full': deploying
-      })}>
+    <div style={{ width: 440 }}>
+      {!isDeploying() && <div className={'w-full h-full bg-white p-8 transition-all duration-300 bg-white p-8 text-left text-black flex flex-col z-10'}>
```
ramfox: it looks like we are losing a transition moment here with this change.

If you look at the deploy screens in `master`, the screen swipes to the left when you click "run deploy", on this branch we don't have a transition.

ramfox (changes requested):
looks like the change makes us lose a transition

b5 (approved):

PR #373, opened 2021-10-06
fix: clearing the dataset state on new dataset redirect

### LINKED ISSUE #369: Title and Name persist when starting a new workflow from an existing one
When clicking "New Dataset" when viewing a workflow, the workflow editor loads as expected, but the dataset name and title remain for the dataset the user started on.  (when scrolling, the mini header says "New dataset from workfflow" as expected, but when scrolled to the top, the header info from the previous dataset is present

boandriy:
clearing the dataset state on new dataset redirect

fixes #369


ramfox (approved):
LGTM!

PR #374, opened 2021-10-06
fix: action dropdown on collections in case of one item in collection will appear on left to the dots

### LINKED ISSUE #366: Dropup menu in collection view appears behind header
image

This dropdown has a `dropUp` flag because the same problem was happening on the last row of the collection view.  Now we are seeing on the first row because when there is only one dataset, it is also the last row.  

I think we need some empty vertical space below the last row (IIRC, we couldn't figure out how to do this using the list component so decided to go with a `dropUp` to change the display of the menu).

boandriy:
since the problem is only visible when there is one ds in collection - 
action dropdown on collections in case of one item in collection will appear on left to the dots : 

image

fixes #366

### DIFF
--- a/src/chrome/DropdownMenu.tsx
+++ b/src/chrome/DropdownMenu.tsx
@@ -10,6 +10,7 @@ interface DropdownMenuProps {
   icon?: string | React.ReactElement
   items: DropdownMenuItemProps[]
   dropUp?: boolean
+  oneItem?: boolean
 }
 
 // DropdownMenu shows menu when a given "icon" prop is clicked. the element to
@@ -20,7 +21,8 @@ const DropdownMenu: React.FC<DropdownMenuProps> = ({
   alignLeft=false,
   className='',
   items,
-  dropUp=false
+  dropUp=false,
+  oneItem=false
 }) => {
   const ref = useRef<HTMLDivElement>(null)
   const [open, setOpen] = useState(false)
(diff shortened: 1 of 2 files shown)


ramfox:
Took a look at the docs, and it looks like if we add the below to the "src/index.css", we don't need to do anything else.
```css
...
/* override the ReactDataTableComponent styles to prevent our drop down from being cut off */
.rdt_TableBody {
  overflow: visible !important;
}
```
unfortunately, this punts the problem a bit. If/when the list of options gets larger, it will get cut off by the `<div>`s that wrap the table. They specify `overflow-y: hidden` currently, which according to the `react-data-table-component` code is necessary because of some specific behavior in firefox:
https://github.com/jbetancur/react-data-table-component/blob/a7315c8a4d041f3841f25707273e623a80e47d85/src/DataTable/ResponsiveWrapper.tsx#L22

but we can solve that when it becomes an issue.

boandriy:
> ```css
> /* override the ReactDataTableComponent styles to prevent our drop down from being cut off */
> .rdt_TableBody {
>   overflow: visible !important;
> }
> ```

Hey @ramfox , this doesn't really solve the issue, + if we give the table the overflow of visible it will lose the scroll.

src/chrome/DropdownMenu.tsx
```diff
@@ -58,7 +65,7 @@ const DropdownMenu: React.FC<DropdownMenuProps> = ({
         role="menu"
         aria-orientation="vertical"
         aria-labelledby="options-menu"
-        style={{ minWidth: '9rem' }}
+        style={{ minWidth: '9rem', ...oneItemStyles}}
```
b5: no need for the `const` declaration above, just use the ternary operator directly:

```ts
style={oneItem
  ? { minWidth: '9rem',  top: -8, right : 30 }
  : { minWidth: '9rem' }
}

I can't remember how the linter likes multiline ternary operators, so please do whatever the linter says is right 😄

b5 (approved):
LGTM!

PR #384, opened 2021-10-07
feat(ci): enable tests

Arqu:
@boandriy tagged you to be aware we're going to add these to the CI, will need to clean up the normal tests and the e2e tests as much as possible.

They're set up not to block deployments for now until we sort out the cloud behavior but will eventually transition to a must pass setup.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,7 +57,7 @@ commands:
 jobs:
   frontend-test:
     docker:
-      - image: circleci/node:14.15
+      - image: circleci/node:15
 
     working_directory: ~/repo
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -86,7 +86,40 @@ jobs:
...
+    - checkout:
+        path: ~/repo
+    - run:
+        name: Check for changes
+        command: ~/repo/.circleci/exit_if_path_not_changed.sh ~/repo
+    - restore_cache:
+        keys:
+          - v1-dependencies-{{ checksum "package.json" }}
+          # fallback to using the latest cache if no exact match is found
+          - v1-dependencies-
```
b5, line 107: neat

b5 (approved):
lgtm!

.circleci/config.yml
Arqu: *cough* I re-used your old code :)

PR #385, opened 2021-10-08
fix: added debounce on ds creation name input

### LINKED ISSUE #364: Sluggish text entry when naming a dataset on deploy
image

I believe this field is validating the name on each keystroke.  When typing quickly, there is a noticeable delay that prevents the text from showing up immediately.

Investigate and see if we can debounce or otherwise make typing into this field happen more smoothly.

boandriy:
added debounce on ds creation name input, so after user stops typing - check for errors

fixes #364

### DIFF
--- a/src/features/workflow/modal/DeployModal.tsx
+++ b/src/features/workflow/modal/DeployModal.tsx
@@ -25,9 +25,9 @@ import { selectDeployStatus } from '../../deploy/state/deployState'
 import { selectSessionUser } from '../../session/state/sessionState'
 import WarningDialog from '../WarningDialog'
 import { removeEvent } from "../../events/state/eventsActions";
+import { useDebounce } from "use-debounce";
 
-
-
+const DEBOUNCE_TIMER = 500
 
 const DeployModal: React.FC = () => {
   const dispatch = useDispatch()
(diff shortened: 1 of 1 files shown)


src/features/workflow/modal/DeployModal.tsx
```diff
@@ -47,8 +48,10 @@ const DeployModal: React.FC = () => {
 
   // local state
   const [ dsName, setDsName ] = useState('')
-  const [ dsNameError, setDsNameError ] = useState()
+  const [debouncedDsName] = useDebounce(dsName, 500)
```
b5: whenever you're using a debounce, declare the interval at a file-level constant just after the block of import statements:

```ts
const debounce_rate = 500
```

That way if we need to tune it later it's visible & clear

src/features/workflow/modal/DeployModal.tsx
```diff
@@ -85,12 +88,14 @@ const DeployModal: React.FC = () => {
...
   const handleDsNameChange = (value: string) => {
-    //validate the dataset name
-    const err = validateDatasetName(value)
-    setDsNameError(err)
     setDsName(value)
   }
 
+  useEffect(() => {
+    if (isNew)
+      setCanBeDeployed(false)
```
b5: always use curly braces on `if` statements

src/features/workflow/modal/DeployModal.tsx
```diff
@@ -103,18 +108,14 @@ const DeployModal: React.FC = () => {
...
-  const checkReadyToDeploy = () => {
-
+  useEffect(() => {
     if (isNew) {
-      // if the name exists and is valid
-      if (!dsName.length || dsNameError) { return false }
+      //validate the dataset name
+      const err = validateDatasetName(debouncedDsName)
+      setDsNameError(err)
+      setCanBeDeployed(!(!debouncedDsName.length || !!err))
```
b5:
```ts
setCanBeDeployed(debouncedDsName.length > 0 && err === null)
```

src/features/workflow/modal/DeployModal.tsx
boandriy: Noted

b5 (approved):
LGTM

PR #388, opened 2021-10-11
fix: displaying status and output for deploy events

### LINKED ISSUE #363: bug(workflow): deploy run output not being shown in the workflow
when we `Deploy` the workflow, the output of the deploy run is not showing in the workflow cells or in the sidebar.

boandriy:
displaying status and output for deploy events, removed deploy events from Toasts

fixes #363

### DIFF
--- a/src/features/deploy/state/deployState.ts
+++ b/src/features/deploy/state/deployState.ts
@@ -2,7 +2,7 @@ import { createReducer } from '@reduxjs/toolkit'
 
 import { RootState } from '../../../store/store'
 import { DeployStatus, Workflow, workflowDeployStatus } from '../../../qrimatic/workflow'
-import { DeployEventAction } from '../../deploy/state/deployActions'
+import { DeployEventAction, ResetRunIdAction } from '../../deploy/state/deployActions'
 import { QriRef } from '../../../qri/ref'
 
 export const DEPLOY_START = 'DEPLOY_START'
@@ -11,6 +11,7 @@ export const DEPLOY_SAVEWORKFLOW_START = 'DEPLOY_SAVEWORKFLOW_START'
 export const DEPLOY_SAVEWORKFLOW_END = 'DEPLOY_SAVEWORKFLOW_END'
 export const DEPLOY_SAVEDATASET_START = 'DEPLOY_SAVEDATASET_START'
 export const DEPLOY_SAVEDATASET_END = 'DEPLOY_SAVEDATASET_END'
+export const DEPLOY_RESET_RUN_ID = 'DEPLOY_RESET_RUN_ID'
 
 
 export function newDeployStatusSelector(workflowID: string): (state: RootState) => DeployStatus {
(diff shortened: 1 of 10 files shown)


src/features/deploy/state/deployState.ts
```diff
@@ -25,13 +25,18 @@ export function selectDeployStatus(qriRef: QriRef): (state: RootState) => Deploy
   return (state: RootState): DeployStatus => (state.deploy.status[`${qriRef.username}/${qriRef.name}`])
 }
 
+export const selectDeployRunId = (state: RootState): string => {
+  return state.deploy.deployRunId
+}
 
 export interface DeployState {
   status: Record<string,DeployStatus>
+  deployRunId: string
```
b5: field with `deploy` prefix on a a structure called `deploy` stutters use `runID` instead, and add a comment:

```ts
export interface DeployState {
  status: Record<string, DeployStatus>
  runID: string  // runID of deploying dataset, empty string means no deploying run
}
```

src/features/workflow/WorkflowPage.tsx
```diff
@@ -21,7 +22,8 @@ const WorkflowPage: React.FC<WorkflowPageProps> = ({ qriRef }) => {
   const dispatch = useDispatch()
   let dataset = useSelector(selectWorkflowDataset)
   const latestDryRunId = useSelector(selectLatestDryRunId)
-  const latestRun = useSelector(selectRun(latestDryRunId))
+  const latestDeployId = useSelector(selectDeployRunId)
+  const latestRun = useSelector(selectRun(latestDryRunId ? latestDryRunId : latestDeployId))
```
b5: rather than defining separate selectors & combining them in this way, we should add a new selector in `workflowState.ts` that pulls from both the deploy _and_ workflow sections of the state tree, combining the `selectDeployRunID` and `selectLatestDryRunID` selectors:

```ts
// workflowState.ts

export const selectLatestRunId = (state: RootState): string => {
  if (state.deploy.runID !== '') { 
    return state.deploy.runID
  }
  if (state.workflow.lastDryRunID) {
    return state.workflow.lastDryRunID
  }
  return ''
}
```

bonus points if you make `workflow.lastDryRunID` non-optional and default to the empty string as "no latest run", as we're doing with `deploy.runID`, then the above can be refactored to just return `state.workflow.lastDryRunID` instead of using the second if clause

src/features/workflow/state/workflowState.ts
```diff
@@ -42,6 +44,13 @@ export const selectLatestDryRunId = (state: RootState): string => {
   return ''
 }
 
+export const selectLatestDryRunDeployId = (state: RootState): string => {
```
b5: should modify this function name to match it's priority, how about `selectLatestDeployOrDryRunId`?

src/features/workflow/state/workflowState.ts
```diff
@@ -92,7 +101,7 @@ export interface WorkflowState {
   // cell output was cleared
   clearedOutputCells: boolean[]
   isDirty: boolean
-  lastDryRunID?: string
+  lastDryRunID: string
```
b5, line 104: ❤️ whenever a field becomes non-optional

src/features/workflow/state/workflowActions.ts
```diff
@@ -18,7 +18,8 @@ import {
   WORKFLOW_MOVE_TRANSFORM_STEP_UP,
   WORKFLOW_MOVE_TRANSFORM_STEP_DOWN,
   WORKFLOW_UNDO_CHANGES,
-  WORKFLOW_CLEAR_TRANSFORM_STEP_OUTPUT
+  WORKFLOW_RESET_DRY_RUN_ID,
+  WORKFLOW_CLEAR_TRANSFORM_STEP_OUTPUT, WORKFLOW_RESET_EDITED_CLEARED_CELLS
```
b5: either go one import per line, or all imports on one line, avoid mixing.

b5 (approved):
a few nits, then looks good to me!

Issue #400, opened 2021-10-12
Data tab is empty when initial navigation is to another tab


chriswhong:
In history view, if the user navigates to another component besides `body` on first render, the "Data" tab is disabled.

http://localhost:3000/chriswhong/fooo/history#structure

boandriy:
@chriswhong , I see that you had a PR that was connected to this issue.  I Can't reproduce this issue with the latest master.

PR #401, opened 2021-10-12
refactor: dataset route refactor, use hash for selected component

### LINKED ISSUE #297: Route Refactoring
Refactor routes to match existing qri cloud

`/preview` ->  `/` 

History should show `/nyc-transit-data/turnstile_daily_counts_2019/history#body`
or `https://rosebud.qri.cloud/nyc-transit-data/turnstile_daily_counts_2019/at/ipfs/QM.../history#body`

No changes to `/workflow`

`/history`-> `/runs`

chriswhong:
- Refactors dataset routes, eliminating `/ds`
  - `preview` no longer appears in the route, and will show at `/:username/:name/`
  -  history is now explicit at `/:username/:name/history#component` or `/:username/:name/at/ipfs/history#component`. This still feels weird as it implies we are looking at the history of a version, not the history of the dataset.  Can revisit later.
  -  Workflow is at `/:username/:name/workflow`
  -  Run log is at `/:username/:name/runs`
- Refactors `ComponentList` to use `Link` for navigation instead of click handlers.  This allows for updating the hash in the URL when selecting a component tab in the history view, and also means the current tab will stay selected when navigating through the commit history!

Closes #297 and #400

### DIFF
--- a/src/features/dataset/DatasetRoutes.tsx
+++ b/src/features/dataset/DatasetRoutes.tsx
@@ -6,7 +6,7 @@
 // DatasetPreviewPage fetches the other necessary parts of the preview (body + readme)
 
 import React, {useEffect} from 'react'
-import { Redirect, Route, Switch, useParams, useRouteMatch } from 'react-router'
+import { Route, Switch, useParams } from 'react-router'
 import { useDispatch } from "react-redux";
 
 import WorkflowPage from '../workflow/WorkflowPage'
@@ -20,7 +20,6 @@ import DatasetWrapper from '../dsComponents/DatasetWrapper'
 import { loadHeader } from "./state/datasetActions";
 
 const DatasetRoutes: React.FC<{}> = () => {
-  const { url } = useRouteMatch()
   // TODO(b5): this qriRef is missing all params after /:username/:name b/c
   // params are dictated by the route that loaded this component.
   // This ref can only be used to load HEAD because DatasetRoutes defines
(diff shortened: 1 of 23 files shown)


src/features/dataset/DatasetRoutes.tsx
```diff
@@ -37,49 +36,42 @@ const DatasetRoutes: React.FC<{}> = () => {
...
+        {/* dataset history */}
+        <Route path='/:username/:name/history'>
+          <DatasetComponents />
         </Route>
 
-        <Route path='/ds/:username/:name/history'>
-          <DatasetActivityFeed qriRef={qriRef} />
+        <Route path='/:username/:name/at/:fs/:hash'>
+          <DatasetComponents />
         </Route>
```
b5: I think the reference-only route should be reserved for the preview page in all cases, and we should add the `/history` suffix. If you're ok with that, copy this up below the other preview routes & add `exact` param

src/features/dataset/state/datasetPaths.ts
```diff
@@ -1,31 +1,32 @@
 import { QriRef } from "../../../qri/ref";
```
b5, line 1: one day we might want to talk about moving these into the routes file. Let's save that chat for later

src/features/dataset/state/datasetPaths.ts
```diff
@@ -1,31 +1,32 @@
...
 
 export function pathToDatasetIssues(ref: QriRef): string {
-  return `/ds/${ref.username}/${ref.name}/issues`
+  return `/${ref.username}/${ref.name}/issues`
 }
 
 export function pathToDatasetPreview(ref: QriRef): string {
-  return `/ds/${ref.username}/${ref.name}/preview`
+  return `/${ref.username}/${ref.name}`
 }
```
b5: expand this to examine the presence of a `path` value in the ref & build the full ref route if so

src/features/dataset/state/datasetPaths.ts
```diff
@@ -1,31 +1,32 @@
...
 
 export function pathToDatasetViewer(ref: QriRef): string {
```
b5: this function is poorly named & kinda weird. It'll direct to history if the ref is present and preview if not. Is that behavior we want? Personally I'd love to see this renamed to `pathToDatasetHistory` and always direct to `/history` endpoints

src/features/dataset/state/datasetPaths.ts
```diff
@@ -1,31 +1,32 @@
...
-  return `/ds/${ref.username}/${ref.name}/edit`
+  return `/${ref.username}/${ref.name}/edit`
 }
 
 export function pathToWorkflowEditor(username: string, name: string): string {
-  return `/ds/${username}/${name}/workflow`
+  return `/${username}/${name}/workflow`
 }
 
 export function pathToActivityFeed(username: string, name:string): string {
```
b5: ditto, I'd love this renamed to `pathToDatasetRuns`

src/features/toast/Toast.tsx
```diff
@@ -28,7 +28,7 @@ const Toast: React.FC<ToastProps> = ({ message, initID, type }) => {
           'text-qrigreen': type === 'succeeded',
           'text-dangerred': type === 'failed',
         })}>{message}</div>
-        <Link to={'/ds/'+ref}
+        <Link to={'/'+ref}
```
b5: template string plz

src/features/workflow/Workflow.tsx
```diff
@@ -57,7 +57,7 @@ const Workflow: React.FC<WorkflowProps> = ({ qriRef }) => {
       dispatch(showModal(ModalType.workflowSplash))
     }
   // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [location])
```
b5, line 60: ah yes this makes sense. we'd need to destructure the location object for this to work properly. We can deal with it later in an audit of hook deps

src/features/workflow/WorkflowDatasetPreview.tsx
```diff
@@ -11,7 +13,7 @@ const WorkflowDatasetPreview: React.FC<WorkflowDatasetPreviewProps> = ({
   dataset
 }) => {
 
-  const [ selectedComponent, setSelectedComponent ] = useState<ComponentName>('body')
+  const selectedComponent = componentFromLocationHash(useLocation())
```
b5: 👌

src/qri/ref.ts
```diff
@@ -73,6 +73,28 @@ export function newQriRef(d: Record<string,any>): QriRef {
   }
 }
 
+export function componentFromLocationHash(location: Record<string,any>) {
```
b5: rename this to `selectorFromLocationHash` and give it's signature an explicit `string` return:

```ts
export function selectorFromLocationHash(location: Record<string,any>): string {
```

the hash component of a route is intended to populate the `selector` field of a `QriRef`, but it needs more thought. Can you add a TODO comment assigned to me that says "this function should be applied to the `selector` field of `QriRef`"?

src/qri/ref.ts
```diff
@@ -73,6 +73,28 @@ export function newQriRef(d: Record<string,any>): QriRef {
...
+    }
+  }
+  return 'body'
+
+}
+
+// parses the selected component from the location hash and appends it to params
+// params can be any object that would normally be passed to newQriRef
+// location is the result of useLocation
+export function mergeParamsAndComponentHash(params: Record<string,any>, location: Record<string,any>): Record<string,any> {
```
b5: how about `refParamsFromLocation`?

src/features/dataset/DatasetRoutes.tsx
chriswhong: Good catch

PR #406, opened 2021-10-13
Cw/various

### LINKED ISSUE #386: User menu should show the username
image

### LINKED ISSUE #389: App chokes when run log is loaded
Navigate to "Run Log" > wait 30 seconds or so > the browser will become unresponsive.

Chrome eventually chokes, giving the option to kill the page.  

Try with this dataset: https://rosebud.qri.cloud/ds/rentsafeto/apartment-building-registration/history

Update: this appears to be an endless loop making calls to `/activity`, check `useEffect()` dependency array

### LINKED ISSUE #390: Short lists should have a trailing border
image

chriswhong:
- fix useEffect loop on run log (closes #389)
- add border to last item in collection and run log (closes #390)
- adds username and email address to SessionUserMenu (closes #38
Home___Qri_Cloud
6)

### DIFF
--- a/src/features/activityFeed/DatasetActivityFeed.tsx
+++ b/src/features/activityFeed/DatasetActivityFeed.tsx
@@ -30,14 +30,8 @@ const DatasetActivityFeed: React.FC<DatasetActivityFeedProps> = ({
 
 
   useEffect(() => {
-    dispatch(loadDatasetLogs(qriRef))
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  },[])
-
-  useEffect(() => {
-    dispatch(loadDatasetLogs(qriRef))
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  },[latestRun?.status])
+    dispatch(loadDatasetLogs({username: qriRef.username, name: qriRef.name}))
+  },[dispatch, qriRef.username, qriRef.name, latestRun?.status])
 
   const handleRunNowClick = () => {
     dispatch(runNow(qriRef))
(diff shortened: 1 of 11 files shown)


src/features/activityFeed/DatasetActivityFeed.tsx
```diff
@@ -31,7 +31,8 @@ const DatasetActivityFeed: React.FC<DatasetActivityFeedProps> = ({
 
   useEffect(() => {
     dispatch(loadDatasetLogs(qriRef))
-  },[dispatch, qriRef])
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
ramfox: okie dokie

so this infinite loop happens because the `qriRef` object changes with every render. the way we have solved this in other places is rather than using `qriRef` in the dependency array, we use `qri.username` and `qri.name`, and any other parts of the ref that we need to respond to. we then create a new `qriRef` to pass into the dispatched method. 

Also, we seem to have two `useEffect` methods that are calling the same function. 

I think we should consolidate like this:

```ts
useEffect(() => {
    dispatch(loadDatasetLogs({username: qriRef.username, name: qriRef.name})
}, [dispatch, qriRef.username, qriRef.name, latestRun?.status]
```

untested code, but that should remove the infinite loop and keep us kosher with the dependency array!

ramfox (changes requested):
one requested change based around our infinite loop/useEffect problem

everything else looks great!

src/features/activityFeed/DatasetActivityFeed.tsx
chriswhong: Thanks, this is super helpful!

PR #407, opened 2021-10-13
fix: removed link to workflow on new dataset

### LINKED ISSUE #399: When making a new workflow, sidebar link contains "undefined"
image

This should link to the current route so that when you click it, nothing happens

boandriy:
removed link to workflow on new dataset

image

fixes #399

### DIFF
--- a/src/features/dataset/DatasetNavSidebar.tsx
+++ b/src/features/dataset/DatasetNavSidebar.tsx
@@ -106,6 +106,7 @@ const DatasetNavSidebar: React.FC<DatasetNavSidebarProps> = ({ qriRef }) => {
           id='workflow-editor'
           icon='code'
           label='Workflow'
+          isLink={!isNewWorkflow}
           to={pathToWorkflowEditor(qriRef.username, qriRef.name)}
           expanded={expanded}
           tooltip={
--- a/src/features/dataset/DatasetSideNavItem.tsx
+++ b/src/features/dataset/DatasetSideNavItem.tsx
@@ -14,6 +14,7 @@ export interface DatasetSideNavItemProps {
   tooltip?: React.ReactNode
   number?: number
   disabled?: boolean
+  isLink?: boolean
 }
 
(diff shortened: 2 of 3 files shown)


src/features/dataset/DatasetNavSidebar.tsx
```diff
@@ -106,7 +106,8 @@ const DatasetNavSidebar: React.FC<DatasetNavSidebarProps> = ({ qriRef }) => {
           id='workflow-editor'
           icon='code'
           label='Workflow'
-          to={pathToWorkflowEditor(qriRef.username, qriRef.name)}
+          isLink={!isNewWorkflow}
+          to={isNewWorkflow ? '/workflow/new' : pathToWorkflowEditor(qriRef.username, qriRef.name)}
```
ramfox: since you've added `isLink`, we don't need to send a `/workflow/new` as a possible path, since `isLink == true` and `to == '/workflow/new'` are mutually exclusive.

Can just be:
```js
    isLink={!isNewWorkflow}
    to={pathToWorkflowEditor(qriRef.username, qriRef.name}
```

ramfox (approved):
LGTM, one small nitpick before we merge though!

src/features/dataset/DatasetNavSidebar.tsx
boandriy: Hey @ramfox , actually the link is used to determine if the text is highlighted (active). And it previously was not active which I think is sort of a bug. Let me know what you think

ramfox: okay i see.

so we should actually make a change here then:

https://github.com/qri-io/frontend/blob/ff9c945e04b1ab56e4a9cd3915068cacaa9cc215/src/features/dataset/state/datasetPaths.ts#L25-L27

if there is no `qriref.name` set, then we can consider this a new workflow, and return `'/workflow/new'`

then in the `DatasetNavSideBar`, we don't have the confusing `!isNewWorkflow` & `isNewWorkflow` logic right next to each other.

PR #408, opened 2021-10-13
fix: removed preview label on ds body if dataset has less then 100 entries

### LINKED ISSUE #282: Indicate that the user is viewing a body preview
Let's add a glyph/text indicator that the body on `/preview` and the body shown when previewing a dry run is a 100 row preview.

boandriy:
removed preview label on ds body if dataset has less then 100 entries

fixes #282

### DIFF
--- a/src/features/dsComponents/body/BodyHeader.tsx
+++ b/src/features/dsComponents/body/BodyHeader.tsx
@@ -24,12 +24,13 @@ const BodyHeader: React.FC<BodyHeaderProps> = ({
 }) => {
   const { structure, body } = dataset
   const headers = extractColumnHeaders(structure, body)
+  const entries = numeral(structure?.entries).format('0,0')
 
   return (
     <div className='flex'>
       <div className='flex flex-grow text-qrigray-400' style={{ fontSize: 12 }}>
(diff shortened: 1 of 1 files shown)


src/features/dsComponents/body/BodyHeader.tsx
```diff
@@ -38,9 +39,9 @@ const BodyHeader: React.FC<BodyHeaderProps> = ({
           TODO (boandriy): Once the preview row count is filtered by fetch pagination -
           use global filtering variable instead of hardcoded 100
         */}
-        <div className={'mr-4'}>
-          Previewing 100 of {numeral(structure?.entries).format('0,0')} rows
-        </div>
+        {+entries > 100 && <div className={'mr-4'}>
```
ramfox: from some in-browser experimenting, it doesn't look like the `+` unary operator works well with string representations of numbers that include commas:

```
var s = "100,000"
+s // NaN
+s > 100 // false
```

Since we specifically get the `entries` variable from using `format('0,0')`, there is a good chance the `entries` string will have a "comma"

Try using `numeral(entries)` rather than `+`

ramfox (approved):
One fix and then it LGTM

PR #409, opened 2021-10-13
feat: added tooltip to user icon on history dataset commit

### LINKED ISSUE #392: Username icon should show username as aa tooltip
image

boandriy:
added tooltip to user icon on history dataset commit

closes #392

### DIFF
--- a/src/chrome/UsernameWithIcon.tsx
+++ b/src/chrome/UsernameWithIcon.tsx
@@ -6,6 +6,7 @@ interface UsernameWithIconProps {
   className?: string
   iconWidth?: number
   iconOnly?: boolean
+  tooltip?: boolean
 }
 
 // TODO(chriswhong): make the prop a user object, or pass in icon URL as a separate prop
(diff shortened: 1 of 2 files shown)


src/chrome/UsernameWithIcon.tsx
```diff
@@ -6,21 +6,23 @@ interface UsernameWithIconProps {
...
   username,
   className,
   iconWidth = 18,
-  iconOnly = false
+  iconOnly = false,
+  tooltip = false
 }) => (
    <div className={classNames('flex items-center tracking-wider', className)}>
-     <div className='rounded-xl inline-block bg-cover flex-shrink-0' style={{
+     <div className='rounded-xl inline-block bg-cover flex-shrink-0' title={tooltip ? username : undefined} style={{
```
ramfox: `title={tooltip ? username : undefined}` => `title={tooltip && username}`

don't use `undefined` as an expression of the ternary operator

ramfox (approved):
One fix that needs to be changed before it can merge. Once that is fixed it LGTM!

src/chrome/UsernameWithIcon.tsx
boandriy: @ramfox , That was my first approach, but had an error with TS:  `Type 'string | false' is not assignable to type 'string | undefined'`

ramfox: then `tooltip ? usename : ''` should work instead

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#title_attribute_inheritance

having the title attribute as the empty string will not show a title

PR #420, opened 2021-10-14
feat: only show collection loading state on first load

chriswhong:
- Shows a loading spinner in place of the list in Collection view *only* if there is no collection state (only on first load)
- If there is already a collection state, show a smaller 'loader' icon next to the page title (this will show when a dataset is removed and the app is refreshing the collection)

When the user removes a dataset from Collection view, the collection list will update in place (the dataset will disappear).  In the future we can make this interaction better by animating the removal.


ramfox (approved):
LGTM

PR #422, opened 2021-10-14
Cw/run now from collection

chriswhong:
Review/Merge #420 first

Adds run now action from collection view.

This introduces a different "Run Now" process from the one we are using in the Run Log.  (Run now from the run log makes use of `workflow` actions and state, which feels wrong, but it's working and I didn't want to break it). 

This approach takes the `runID` returned from the API call to `/auto/run` and mutates the value stored for that dataset in `collection` state.  That `runID` is then used by `ManualTriggerButton` to show a loading spinner in the correct row in collection view.

- Adds some logic to `ManualTriggerButton` to refresh the collection state when the run is successful.
- Restores the "automation" icon in collection view based on the existence of `runID`
- Some tweaks/fixes to various internal dataset links

### DIFF
--- a/src/features/activityFeed/ActivityList.tsx
+++ b/src/features/activityFeed/ActivityList.tsx
@@ -76,7 +76,7 @@ const ActivityList: React.FC<ActivityListProps> = ({
           }
         })
         if (!['failed', 'unchanged'].includes(row.runStatus)) {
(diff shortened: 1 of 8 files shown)


ramfox (approved):
LGTM. There's some weird behavior when tracking the last run time & run count, but those will need changes from the backend & this looks good so I think we should go ahead and merge & fix in a follow up PR.

PR #424, opened 2021-10-15
fix: displaying dataset title on workflow header

### LINKED ISSUE #411: Title/header missing on new workflow
image

We show "New dataset from workflow" in the mini header after the user scrolls, but there is no title on the workflow editor when the user hasn't scrolled yet.

boandriy:
displaying dataset title on workflow header

fixes #411

### DIFF
--- a/src/features/dataset/DatasetHeader.tsx
+++ b/src/features/dataset/DatasetHeader.tsx
@@ -14,20 +14,21 @@ import fileSize from "../../utils/fileSize";
 
 
 export interface DatasetHeaderProps {
+  isNew: boolean
   border?: boolean
   editable?: boolean
-  showInfo?: boolean
 }
 
+export const newWorkflowTitle: string = 'New Dataset from Workflow'
 // DatasetHeader and DatasetMiniHeader now accept children which will be displayed
 // in the right side where the default buttons are located. This is useful for
 // overriding the download button with the dry run button in the workflow editor
 
 const DatasetHeader: React.FC<DatasetHeaderProps> = ({
+  isNew,
   border = false,
   editable = false,
-  showInfo = true,
-  children
+  children,
 }) => {
   const dispatch = useDispatch()
   const history = useHistory()
@@ -50,7 +51,7 @@ const DatasetHeader: React.FC<DatasetHeaderProps> = ({
       <div className='flex mb-5'>
         <div className='flex-grow'>
           {/* don't show the username/name when creating a new dataset with the workflow editor */}
-          { qriRef.username && (
+          { !isNew && (
             <div className='text-md text-gray-400 relative flex items-center group hover:text pb-1 font-mono h-10'>
               <Link to={`/${qriRef.username}`} className='whitespace-nowrap' colorClassName='text-qrigray-400 hover:text-qrigray-800'>{qriRef.username}</Link>/
               <EditableLabel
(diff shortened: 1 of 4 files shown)


ramfox (changes requested):
So this works, but I think we can use this opportunity to do a bit of clean up around `DatasetHeader`, `DatasetMiniHeader`, etc

This is a long comment requesting a lot of changes so please read it carefully

## `isNew`
We keep needing to determine if the workflow `isNew` from `WorkflowPage` => `DatasetScrollLayout => `DatasetHeader` & `DatasetMiniHeader`

Let's calculate this in one place, that way if anything changes about the way we need to calculate `isNew`, there will be only one place we need to alter it.

- [x] Let's add an `isNew` params to `DatasetScrollLayout`, `DatasetMiniHeader` and `DatasetHeader`, and pass `isNew` down from `WorkflowPage`.

## unneeded `dataset` object

In `WorkflowPage`, if a workflow `isNew` we edit the `dataset` object so that it can be used in `DatasetHeader` and `DatasetMiniHeader`. However, this looks like "hold over" from when we were using the `dataset` object to populate the header, we now use the `header` object to populate those.

- [x] instead, let's remove these lines:
https://github.com/qri-io/frontend/blob/796078355dda7abfef5f9b5ba05810861e8950b4/src/features/workflow/WorkflowPage.tsx#L30-L39

- [x] Let'a also remove the `dataset` field from the `DatasetScrollLayoutProps`, `DatasetMiniHeaderProps`, & `DatasetHeader` props.

- [x] Instead, in `DatasetHeader` let's create a `const newWorkflowTitle: string = 'New Dataset from Workflow'`

- [x] In `DatasetMiniHeader` & `DatasetHeader`, use the `isNew` param to determine if we should show the `newWorkflowTitle` or the `header.name` (import `newWorkflowTitle` into `DatasetMiniHeader` from `DatasetHeader`)

## `DatasetHeader` further refactors
We have a few places where we are relying on `qriRef.username` or `qri.name` to determine if we are showing or hiding certain pieces of the UI. Let's instead rely on the new `isNew` param.

- [x] First, let's remove the `showInfo` param & replace it with `isNew`. 

- [x] Second, let's replace the check for `qri.username` with a check for the new `isNew` param:
https://github.com/qri-io/frontend/blob/796078355dda7abfef5f9b5ba05810861e8950b4/src/features/dataset/DatasetHeader.tsx#L52-L53

## simplify logic in `WorkflowPage`
In `WorkflowPage`, we have this bit of code to show if we should be showing a loading screen:
https://github.com/qri-io/frontend/blob/796078355dda7abfef5f9b5ba05810861e8950b4/src/features/workflow/WorkflowPage.tsx#L52-L60

- [x] I personally find this harder to parse then if we just switched some of the logic:

```js
{dataset || isNew
      ? (
        <DatasetScrollLayout headerChildren={runBar} useScroller>
          <Workflow qriRef={qriRef} />
        </DatasetScrollLayout>
    )
      : (<div className='w-full h-full p-4 flex justify-center items-center'>
      <Spinner color='#43B3B2' />
    </div>)}
```

-----

There may be more small changes, but the main point is to ensure that we aren't passing around the `dataset` object where we don't actually use it & to have only one place where we are calculating `isNew`, so that it is easier to refactor in the future!

lmk if you have questions!

ramfox (approved):
LGTM

PR #426, opened 2021-10-15
fix: added logic for workflow links active state

### LINKED ISSUE #425: Only one link should be active on workflow sidebar
image

boandriy:
added logic for workflow links active state

fixes #425

### DIFF
--- a/src/features/dataset/DatasetSideNavItem.tsx
+++ b/src/features/dataset/DatasetSideNavItem.tsx
@@ -15,6 +15,7 @@ export interface DatasetSideNavItemProps {
   number?: number
   disabled?: boolean
   isLink?: boolean
+  exact?: boolean
 }
 
 const DatasetSideNavItem: React.FC<DatasetSideNavItemProps> = ({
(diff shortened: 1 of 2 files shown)


src/features/dataset/DatasetSideNavItem.tsx
```diff
@@ -26,10 +27,19 @@ const DatasetSideNavItem: React.FC<DatasetSideNavItemProps> = ({
...
+  isLink = true,
+  exact = false,
 }) => {
   const { pathname } = useLocation();
-  const active = pathname.includes(to)
+
+  let active
+  if (exact) {
+    active = pathname === to
+  } else if (to.includes('history') && pathname.includes('history')) {
```
boandriy: Hey @ramfox , @b5 , would really like your thoughts on how to improve this logic. Since we have the link to history as `..../history#body` - pathname doesn't include the `#body` in it. So the previous logic was always returning false for the history link. Furthermore, the links `to` is always the same, meaning that if we would go to a specific commit - the logic would also not work since the route is now including all the commit info in it and links `to` property does not.

ramfox: okay great. yea this is tricky!

Our main edge case is - what if someone's username or name includes the text 'history'?

I think this is the annoying thing we have to do to cover all our weird cases:

create an `isActive` function, that passes in the `to` & `pathname` variables

- Split the strings on `#`, discarding everything after the `#`
- If `pathname.includes(to)` return true
- The only case left to test should be the `history` case
  - split both strings on `/`
  - if there are only two elements return false
  - first element of both arrays should match
  - 2nd element of both arrays should match
  - last element of both arrays should match 'history'

ramfox (approved):
LGTM

PR #427, opened 2021-10-15
fix: modified e2e tests to support new routes names

boandriy:
modified e2e tests to support new routes names

Hey, @b5 , @chriswhong , @ramfox , From now on we need to do the e2e runs after each change to the FE, to make sure that new changes didn't break the tests. Should we document it somewhere?


ramfox (approved):
LGTM

To answer your question, yes we should. I believe it will soon be a "required" part in CI

PR #428, opened 2021-10-15
fix: moved the preview ds body label to existing row count label

### LINKED ISSUE #413: Preview message is redundant
image

This is redundant... let's just use the rows icon and then say either:
"Previewing 100 of 37,234 rows" or "Showing all 47 rows"

boandriy:
moved the preview ds body label to existing row count label

Preview:
image
All dataset: 
image

fixes #413


ramfox (approved):
LGTM

PR #434, opened 2021-10-15
feat: add tooltips where we use the automation icon

chriswhong:
- Adds browser-built-in tooltips wherever we use the small "automated" icon (in commit list, collection, and run log).
- Adds browser-built-in tooltip to the large automation icon that appears in the left-most column of collection view.


src/chrome/DatasetCommitInfo.tsx
```diff
@@ -24,7 +24,7 @@ const DatasetCommitInfo: React.FC<DatasetCommitInfoProps> = ({
     })}>
       <div className={classNames('text-black font-semibold flex justify-between items-center mb-2')}>
         <div className='dataset_commit_info_text'>{dataset.commit?.title}</div>
-        <div className='flex-grow-0 text-qrigreen'>
+        <div className='flex-grow-0 text-qrigreen' title={'version created by this dataset\'s transform script'}>
```
b5: drop braces & just use a string literal

b5 (approved):
one nit then LGTM

PR #435, opened 2021-10-16
feat(server): simple express app to host our react webapp

ramfox:
Two new commands:

`yarn serve-dev`: this will serve the express app using `nodemon` which will pick up on any changes to the express app as you are developing. It expects that the react app has already been built.

`yarn serve`: this will build the production version of the app and serve it using `node`

The express app will also listen for url requests that match `/:username/:name`, and request the associated dataset preview for that dataset. It will craft html tags that will embed the dataset preview data into the html DOM in such a way that it can be picked up by google's dataset search.

----
This also alters the dockerfile, using the new `yarn serve` command to build & serve the app.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 
 # dependencies
 /node_modules
+/server/node_modules
 /.pnp
 .pnp.js
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,7 @@
 # See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
 
 # dependencies
-/node_modules
-/server/node_modules
+node_modules
 /.pnp
 .pnp.js
 
--- a/server/index.js
+++ b/server/index.js
@@ -1,62 +1,96 @@
 const express = require("express")
 const path = require('path')
+const fs = require('fs')
+const fetch = require('node-fetch')
+const composeHeadData = require('./util').composeHeadData
+const composeJSONLD = require('./util').composeJSONLD
+
+const API_BASE_URL = process.env.REACT_APP_API_BASE_URL || 'http://localhost:2503'
+const PORT = 3000 
+
 const app = express()
-const PORT = 3001 
 
 app.use(express.static(path.join(__dirname, "..", "build")))
 app.use(express.static(path.join(__dirname, "..", "public")))
 
-app.listen(PORT, () => {
+var server = app.listen(PORT, () => {
   console.log(`server started on port ${PORT}`)
 })
 
-// const passThroughRoutes = [
-//   '/login',
-//   '/signup',
-//   '/login/forgot',
-//   '/dashboard',
-//   '/collection',
-//   '/activity',
-//   '/workflow/new',
-//   '/new',
-//   '/run',
-//   '/changes',
-//   '/search',
-//   '/notifications',
-//   '/notification_settings',
-//   '/:username',
-//   '/:username/following'
-// ]
+// reservedUsernames are a list of words that may be mistaken for usernames
+// during a url regex match of `/:username/:name`, because of how our routes
+// are formed
+// must be updated if routes change, particularly routes with 2 segments as
+// those are the routes at risk for being incorrectly matched
+const reservedUsernames = ['ipfs', 'login', 'workflow'] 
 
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -2,6 +2,7 @@
 
 # dependencies
 /node_modules
+/server/node_modules
```
b5: change this to just `node_modules`, it'll ignore everywhere

server/index.js
```diff
@@ -0,0 +1,99 @@
...
+// are formed
+// must be updated if routes change, particularly routes with 2 segments as
+// those are the routes at risk for being incorrectly matched
+const reservedNames = ['following']
+
+const indexPath = path.join(__dirname, "..", "build", "index.html")
+
+fs.access(indexPath, fs.constants.F_OK, (err) => {
+  if (err !== null) {
+    console.log(`File ${indexPath} does not exists. App must be built before attempting to host it.`)
```
b5: bump this up to `console.error`, add the err itself to the log message, change `exists` -> `exist`

server/index.js
```diff
@@ -0,0 +1,99 @@
...
+    server.close(() => {
+      process.exit(1)
+    }) 
+  }
+})
+
+app.use('/', async (req, res) => {
```
b5: this'll work for now. Can I get a `TODO(ramfox): separate routes into a base route and "/:username/:name"`?

server/index.js
```diff
@@ -0,0 +1,99 @@
+const express = require("express")
+const path = require('path')
+const fs = require('fs')
+const match = require('path-to-regexp').match
+const fetch = require('node-fetch')
+const composeHeadData = require('./util').composeHeadData
+const composeJSONLD = require('./util').composeJSONLD
+
+const API_BASE_URL = process.env.REACT_APP_API_BASE_URL || 'http://localhost:2503'
```
b5, line 8: we shouldn't use `:2503` as the default port here, that's the qri API server route. how about defaulting to `:3000`? That's what the frontend HMR server runs on

ramfox: We do run on port 3000, this is just the to set up the base url of the API server we fetch from

server/util.js
```diff
@@ -0,0 +1,50 @@
+const composeHeadData = (dataset) => {
+  const { peername, name, meta } = dataset
+  // start with generic title and description
+  let title = `${peername}/${name} | Dataset Published on qri.cloud`
```
b5: for these, use github's titles as a guideline. these additions don't do much for SEO these days, and gets cut off by everyone's giant list of tabs. With that said, "qri" is definitely a newer tool than github, so let's add `| qri.cloud` to the end so folks know where they are:

```ts
let title = `${peername}/${name} | qri.cloud`
```

for branding purpose always use `qri.cloud`, not `Qri Cloud`

server/util.js
```diff
@@ -0,0 +1,50 @@
...
+  let description = `Preview this dataset on Qri Cloud`
```
b5: `Preview this dataset on qri.cloud`

server/util.js
```diff
@@ -0,0 +1,50 @@
...
+    peername,
+    name,
+    meta = {}
+  } = dataset
+
+  const jld = {
+    '@context': 'https://schema.org/',
+    '@type': 'Dataset',
+    name: meta.title || name,
+    description: meta.description || `A dataset published on Qri Cloud by user ${peername}`,
```
b5: drop the `user`:

```ts
meta.description || `A dataset published on qri.cloud by ${peername}`
```

server/util.js
```diff
@@ -0,0 +1,50 @@
+const composeHeadData = (dataset) => {
+  const { peername, name, meta } = dataset
+  // start with generic title and description
+  let title = `${peername}/${name} | Dataset Published on qri.cloud`
+  let description = `Preview this dataset on Qri Cloud`
+
+  // if meta, use meta values
+  if (meta) {
+    if (meta.title) {
+      title = `${meta.title} | Dataset Published on qri.cloud`
```
b5: match above comment:
`| Dataset Published on qri.cloud` -> `| qri.cloud`

b5 (approved):
looks great, just some presentation nits

PR #438, opened 2021-10-19
refactor: standardize Button, add MiniButton

chriswhong:
Refactors `Button` to make all buttons site-wide match the types and sizes in our style guide.
- Adds `icon` prop for standardized placement and sizing of icons before button text
- Adds `block`prop for full-width buttons with larger border radius
- Adds `caret` prop for showing a small caret to the right of the button text

Adds a separate component `MiniButton` for icon-only square and round buttons (as used in the cron scheduler and the Add Dataset button in the collapsed dataset menu.)


src/chrome/MiniButton.tsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react'
+import classNames from 'classnames'
+
+import Icon, { IconSize } from './Icon'
+
```
b5: extraneous newline

src/chrome/MiniButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { ButtonType, generateButtonTypeClasses } from './Button'
+
+export interface MiniButtonProps {
```
b5: what about calling this `IconOnlyButton`, since it's a button that can't have text? Feels strange to have a `MiniButton` with a `size='lg'` prop

src/chrome/MiniButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+  // icon is required
+  icon: string
```
b5: this type should really be an enum of all possible icon strings, but that isn't defined in the `Icon` component yet. I'll file an issue

src/chrome/MiniButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+    dimension = 36
+    iconSize = 'sm'
+    borderRadius = round ? 78.3 : 6.26
+  }
+
+  if (size === 'lg') {
+    dimension = 46
+    iconSize = 'md'
+    borderRadius = round ? 100 : 8
+  }
```
b5: group these with an `else if`

src/chrome/Button.tsx
```diff
@@ -10,59 +12,130 @@ export type ButtonType = 'primary' // light blue button with white text
...
+
+    if (size === 'md') {
+      iconSize = 'sm'
+      height = 38
+    }
+
+    if (size === 'lg') {
+      iconSize = 'md'
+      height = 44
+    }
```
b5: group these with an `else if`, move `height = 23` into an `if (size === 'sm')` block so height is only assigned once

src/chrome/Button.tsx
```diff
@@ -10,59 +12,130 @@ export type ButtonType = 'primary' // light blue button with white text
...
+    height = 26
+    sizeClassName = 'w-full rounded-xl'
+
+    if (size === 'md') {
+      height = 36
+    }
+
+    if (size === 'lg') {
+      height = 46
+    }
```
b5: group these with an `else if`, move first two assignments into an `if (size === 'sm)` block

src/chrome/Button.tsx
```diff
@@ -10,59 +12,130 @@ export type ButtonType = 'primary' // light blue button with white text
...
+    sizeClassName = 'rounded-md px-2.5'
+    caretSize = '2xs'
+  }
+
+  if (size === 'lg') {
+    fontSize = 14
+    height = 38
+    sizeClassName = 'rounded-lg px-8'
+    caretSize = 'xs'
+  }
```
b5: group these with an `else if`. default the above to `md` settings (which is the default size prop), and flip the `if` conditional to check if `size === 'sm'`. This'll make fewer assignments in the default case

b5 (approved):
a few nits, then LGTM

PR #439, opened 2021-10-19
refactor(RunBar): disable `DeployButton` if there is no session

### LINKED ISSUE #277: Workflow API call errors when logged out
Go to http://localhost:3000/ds/chriswhong/brooklyn_hourly_temperature_humidity/workflow when not logged in, the transform script does not appear on the page

While we are at it:
- [x] Disable Deploy button on existing workflow when not logged in

ramfox:
closes #277

### DIFF
--- a/src/features/workflow/state/workflowActions.ts
+++ b/src/features/workflow/state/workflowActions.ts
@@ -10,8 +10,6 @@ import {
   SET_TEMPLATE,
   SET_WORKFLOW,
   SET_WORKFLOW_REF,
-  SET_RUN_MODE,
-  RunMode,
   WORKFLOW_ADD_TRANSFORM_STEP,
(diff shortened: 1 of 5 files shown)


src/features/workflow/state/workflowActions.ts
```diff
@@ -11,7 +11,6 @@ import {
   SET_WORKFLOW,
   SET_WORKFLOW_REF,
   SET_RUN_MODE,
```
boandriy: @ramfox , could you also delete the unused `SET_RUN_MODE`

boandriy (approved):
LGTM!

src/features/workflow/state/workflowActions.ts
ramfox: good catch!!!

PR #444, opened 2021-10-20
fix(ci): ensure CI waits for deploy for e2e tests

Arqu:


b5 (approved):
LGTM. I'm increasingly thinking we should build an isolated, ephemeral environment for running e2e tests again (like we had with desktop), but that's at least a few months off pending more cloud cleanup work. This works well in the meantime

Arqu:
I've got ideas on that front, but it's a fight for another day.

PR #445, opened 2021-10-20
feat: adding history loading state

### LINKED ISSUE #382: History View > Loading State and Transition between commits
Instead of a spinner, the body shows "There's nothing here yet" until the body loads.  A spinner would be good, but maybe we can show a grayed out table.

boandriy:
adding history loading state

closes #382

### DIFF
--- a/src/features/commits/CommitSummaryHeader.tsx
+++ b/src/features/commits/CommitSummaryHeader.tsx
@@ -1,32 +1,48 @@
 import React from 'react'
+import ContentLoader from "react-content-loader";
 
 import Dataset from '../../qri/dataset'
 import DatasetVersionInfo from '../../chrome/DatasetCommitInfo'
 
+
 export interface CommitSummaryHeaderProps {
   dataset: Dataset
+  loading?: boolean
 }
 
 const CommitSummaryHeader: React.FC<CommitSummaryHeaderProps> = ({
   dataset,
+  loading = false,
   children
 }) => {
   const { commit } = dataset
-  if (commit) {
+    if (!commit && !loading) {
+      return null
+    }
+
     return (
       <div className='min-height-200 py-4 px-8 rounded-lg bg-white flex'>
         <div className='commit_summary_header_container'>
-          <div className='text-sm text-qrigray-400 font-semibold mb-2'>Version Info</div>
-          <DatasetVersionInfo dataset={dataset} />
+          { loading ?
+            <>
+              <div className='text-sm text-qrigray-400 font-semibold mb-2'>Version Info</div>
+              <ContentLoader width="421" height="48.6"  fill="none">
+                <rect y="4" width="421" height="12" fill="#D5DADD"/>
+                <rect y="32" width="89" height="12"  fill="#D5DADD"/>
+                <rect x="99" y="32" width="89" height="12" fill="#D5DADD"/>
+              </ContentLoader>
+            </>:
+            <>
+              <div className='text-sm text-qrigray-400 font-semibold mb-2'>Version Info</div>
+              <DatasetVersionInfo dataset={dataset} />
+            </>
+          }
         </div>
         <div className='flex-grow flex items-center justify-end'>
           {children}
         </div>
       </div>
     )
-  }
-
-  return null
 }
 
 
--- a/src/features/commits/DatasetCommit.tsx
+++ b/src/features/commits/DatasetCommit.tsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import { Link, useLocation } from 'react-router-dom'
 import classNames from 'classnames'
+import ContentLoader from "react-content-loader";
 
 import { LogItem } from '../../qri/log'
 import { newQriRef, refParamsFromLocation } from '../../qri/ref'
@@ -11,12 +12,14 @@ export interface DatasetCommitProps {
   logItem: LogItem
   active?: boolean
   isLink?: boolean
+  loading?: boolean
 }
 
 const DatasetCommit: React.FC<DatasetCommitProps> = ({
   logItem,
   active = false,
-  isLink = true
+  isLink = true,
+  loading = false
 }) => {
   const location = useLocation()
   // create a Dataset to pass into DatasetCommitInfo
@@ -30,13 +33,20 @@ const DatasetCommit: React.FC<DatasetCommitProps> = ({
     runID: logItem.runID
   }
 
-  const content = (
+  const content = loading ?
+    (<ContentLoader height='38.6' width='177.5'>
+      <rect width="145" height="11" y='2' fill="#D5DADD"/>
+
+      <rect width="20" y='24' height="11" rx="1" fill="#D5DADD"/>
+      <rect x="26" y='24' width="30" height="11" rx="1" fill="#D5DADD"/>
+      <rect x="66" y='24' width="60" height="11" rx="1" fill="#D5DADD"/>
+    </ContentLoader>) : (
     <DatasetCommitInfo dataset={dataset} small automated={!!dataset.runID} />
   )
 
   const containerClassNames = classNames('block rounded-md px-3 py-3 mb-6 w-full overflow-x-hidden', active && 'bg-white', !active && 'text-qrigray-400 border border-qrigray-300')
 
-  if (isLink) {
+  if (isLink && !loading) {
     return (
       <Link
         className={containerClassNames}
--- a/src/features/commits/DatasetCommitListItem.tsx
+++ b/src/features/commits/DatasetCommitListItem.tsx
@@ -6,6 +6,7 @@ import DatasetCommit from './DatasetCommit'
 
 export interface DatasetCommitListItemProps {
   logItem: LogItem
+  loading?: boolean
   active?: boolean
   first?: boolean
   last?: boolean
@@ -13,6 +14,7 @@ export interface DatasetCommitListItemProps {
 
 const DatasetCommitListItem: React.FC<DatasetCommitListItemProps> = ({
   logItem,
+  loading = false,
   active = false,
   first = false,
   last = false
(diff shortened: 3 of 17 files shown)


chriswhong (changes requested):
- The loading content for `CommitSummaryHeader` isn't the same height as the content, so there is a shift that happens when it appears and disappears.  Let's make sure this element has a constant height
- The tab for "Data" is wider than the other tabs by design, but during the loading state, it's the same width.  Let's make sure the Data tab is always the same width and is wider than the others.
- The body still displays "There's nothing here yet" during part of the loading... When I tested it I saw the new table with animated loading rectangles in it, but this went away, then it shows "There's nothing here yet", then the body renders.  Let's make sure the loading state table shows continuously until the body is in place.

boandriy:
Hey @chriswhong , There is still a content shift, as the header is missing on the loading state. Should I add similar content loader?

chriswhong (changes requested):
Still seeing a very slight shift in content when switching between versions in history.  Also, the heading Version Info can stay as it doesn't change, only the content below it changes.

Can we resize/reformat the SVG so that it matches the height of the content it is replacing and prevents the shift completely?

chriswhong:
A few more changes:
- The active tab shouldn't be changing either.  I am seeing the data tab as white, then choosing another version, and it flashes gray and then white again.  Let's make sure this flash doesn't occur, the active tab should stay active through the loading state.
- I see "showing all 0 rows" for a moment which is awkward.  I think because these body header icons and text are derived from the content that is loading, they should also show loading state.

src/features/dataset/state/datasetState.ts
```diff
@@ -78,8 +84,12 @@ export const datasetReducer = createReducer(initialState, {
   'API_DATASET_FAILURE': (state) => {
     state.datasetLoading = false
   },
+  'API_BODY_REQUEST': (state, action) => {
+    state.bodyLoading = true
+  },
   'API_BODY_SUCCESS': (state, action) => {
     state.dataset.body = action.payload.data as Body
+    state.bodyLoading = false
```
ramfox, line 94: need to also track `API_BODY_FAILURE`, and set the `bodyLoading` to false

src/features/dsComponents/body/Body.tsx
```diff
@@ -35,6 +36,8 @@ export interface BodyProps {
 // }
 
 const extractColumnHeaders = (structure: Structure, value: any[]): ColumnProperties[] => {
+
```
ramfox: extra line here

src/features/dsComponents/body/Body.tsx
```diff
@@ -1,12 +1,13 @@
...
-import { useDispatch } from 'react-redux'
+import {useDispatch, useSelector} from 'react-redux'
 // import { Action } from 'redux'
 
 import Dataset,  { Structure, schemaToColumns, ColumnProperties } from '../../../qri/dataset'
 import BodyTable from './BodyTable'
 import BodyJson from './BodyJson'
 import { loadBody } from '../../dataset/state/datasetActions'
 import { newQriRef } from '../../../qri/ref'
+import {selectIsBodyLoading} from "../../dataset/state/datasetState";
```
ramfox:
```
import {selectIsBodyLoading} from "../../dataset/state/datasetState";
```
=>
```
import { selectIsBodyLoading } from "../../dataset/state/datasetState"
```

src/features/dsComponents/body/Body.tsx
```diff
@@ -1,12 +1,13 @@
 import React, { useEffect } from 'react'
-import { useDispatch } from 'react-redux'
+import {useDispatch, useSelector} from 'react-redux'
```
ramfox:
```
import {useDispatch, useSelector} from 'react-redux'
```
=>
```
import { useDispatch, useSelector } from 'react-redux'
```

src/features/dsComponents/body/BodyTable.tsx
```diff
@@ -14,6 +15,7 @@ import DataType from '../../../chrome/DataType'
 interface BodyTableProps {
   headers?: any[]
   body: any[]
+  loading: boolean
```
ramfox: can we switch this to optional? Then default to `false` when you destructure it from `props`? Then we can remove `loading={false}` from the `<BodyTable>` call in `Body.tsx`

src/features/dsComponents/body/Body.tsx
```diff
@@ -125,6 +138,7 @@ const Body: React.FC<BodyProps> = ({
   return (structure.format === 'csv' && Array.isArray(body))
     ? <BodyTable
         headers={extractColumnHeaders(structure, body)}
+        loading={false}
```
ramfox: Can we make the `loading` param optional (defaulting to `false`), then remove this line?

ramfox (approved):
A few nitpicks, but this looks great!

src/features/commits/CommitSummaryHeader.tsx
```diff
@@ -2,31 +2,46 @@ import React from 'react'
 
 import Dataset from '../../qri/dataset'
 import DatasetVersionInfo from '../../chrome/DatasetCommitInfo'
+import ContentLoader from "react-content-loader";
```
b5: move package imports above local imports, separated by a blank linke

src/features/commits/CommitSummaryHeader.tsx
```diff
@@ -2,31 +2,46 @@ import React from 'react'
...
 
 export interface CommitSummaryHeaderProps {
   dataset: Dataset
+  loading: boolean
```
b5: is this prop _required_? we often do loading props as optional with a default value of `false` within the component itself

src/features/commits/CommitSummaryHeader.tsx
```diff
@@ -2,31 +2,46 @@ import React from 'react'
...
+              <div className='text-sm text-qrigray-400 font-semibold mb-2'>Version Info</div>
+              <ContentLoader width="421" height="48.6"  fill="none">
+                <rect y="4" width="421" height="12" fill="#D5DADD"/>
+                <rect y="32" width="89" height="12"  fill="#D5DADD"/>
+                <rect x="99" y="32" width="89" height="12" fill="#D5DADD"/>
+              </ContentLoader>
+            </>:
+            <>
+              <div className='text-sm text-qrigray-400 font-semibold mb-2'>Version Info</div>
+              < DatasetVersionInfo dataset={dataset} />
```
b5: extraneous space: `< DatasetVersion...`

src/features/commits/DatasetCommit.tsx
```diff
@@ -6,17 +6,20 @@ import { LogItem } from '../../qri/log'
 import { newQriRef, refParamsFromLocation } from '../../qri/ref'
 import { pathToDatasetHistory } from '../dataset/state/datasetPaths'
 import DatasetCommitInfo from '../../chrome/DatasetCommitInfo'
+import ContentLoader from "react-content-loader";
```
b5: move package imports above local imports

src/features/commits/DatasetCommitListItem.tsx
```diff
@@ -6,13 +6,15 @@ import DatasetCommit from './DatasetCommit'
 
 export interface DatasetCommitListItemProps {
   logItem: LogItem
+  loading: boolean
```
b5: make this optional, with a default value of `false`

src/features/dataset/DatasetHeader.tsx
```diff
@@ -8,9 +8,10 @@ import DatasetInfoItem from './DatasetInfoItem'
 import DownloadDatasetButton from '../download/DownloadDatasetButton'
 import Link from '../../chrome/Link'
 import { validateDatasetName } from '../session/state/formValidation'
-import { selectDatasetHeader } from "./state/datasetState";
+import { selectDatasetHeader, selectIsHeaderLoading } from "./state/datasetState";
 import { qriRefFromVersionInfo } from "../../qri/versionInfo";
 import fileSize from "../../utils/fileSize";
+import ContentLoader from "react-content-loader";
```
b5: move package import above local import

src/features/dsComponents/body/BodyTable.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react'
 import classNames from 'classnames'
+import ContentLoader from "react-content-loader";
 // import * as _ from 'underscore'
 // import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 // import { faSync } from '@fortawesome/free-solid-svg-icons'
```
b5: remove these commented-out imports

src/features/dsComponents/body/Body.tsx
```diff
@@ -60,10 +62,20 @@ const Body: React.FC<BodyProps> = ({
...
+    }
   }, [preview, dispatch, path, name, username])
 
-  if (!body) {
+  if (loading) {
+    return <BodyTable
+      headers={[]}
+      loading={true}
+      body={[]}
+    />
```
b5: make `headers` and `body` BodyTable props optional, use an empty array as default values. use `loading` prop literal instead:

```ts
if (loading) {
  return <BodyTable loading />
}
```

src/features/dsComponents/DatasetComponents.tsx
```diff
@@ -21,7 +21,10 @@ const DatasetComponents: React.FC<{}> = () => {
 
   useEffect(() => {
     const ref = newQriRef({username: qriRef.username, name: qriRef.name, path: qriRef.path})
-    dispatch(loadDataset(ref))
+    dispatch(setBodyLoading())
+    if (ref.username && ref.name) {
+      dispatch(loadDataset(ref))
+    }
```
b5, line 27: This solution works for now, but isn't optimal. Body loading state should be handled entirely within the `loadDataset` action, and a selector like `selectBodyIsLoading` should be able to work out it's state entirely from the request lifecycle of `loadDataset`.

That said, this solution will work for now, so let's merge it. Can you file an issue titled `Revisit Body Loading State on Dataset Components page` and quote my above comment?

b5 (approved):
LGTM

PR #446, opened 2021-10-20
fix: disabling run log link if no workflow

### LINKED ISSUE #414: Run Log should be disabled unless workflow exists
image

boandriy:
disabling run log link if no workflow

fixes #414


chriswhong:
LGTM, I pushed a commit that removes `cursor-pointer` when the item is disabled.

chriswhong (approved):

b5 (approved):
LGTM

PR #447, opened 2021-10-20
fix: run log commit message replaced with title

### LINKED ISSUE #437: Commit Title mismatched between history and run log
Home___Qri_Cloud
image

Observed two references to the same commit, but the titles appear to be different between the history view and the run log

boandriy:
run log commit message replaced with title to match commit message on history page

fixes #437


ramfox (approved):
LGTM

PR #449, opened 2021-10-20
refactor(server): get `PORT` from env var

ramfox:
Also adding more potentially "reservedUsernames" that may conflict with
app routes

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -6,7 +6,7 @@ const composeHeadData = require('./util').composeHeadData
 const composeJSONLD = require('./util').composeJSONLD
 
 const API_BASE_URL = process.env.REACT_APP_API_BASE_URL || 'http://localhost:2503'
-const PORT = 3000 
+const PORT = process.env.PORT || 3000 
 
 const app = express()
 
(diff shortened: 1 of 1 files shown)


server/index.js
```diff
@@ -22,7 +22,7 @@ var server = app.listen(PORT, () => {
 // are formed
 // must be updated if routes change, particularly routes with 2 segments as
 // those are the routes at risk for being incorrectly matched
-const reservedUsernames = ['ipfs', 'login', 'workflow'] 
+const reservedUsernames = ['ipfs', 'login', 'workflow', 'admin', 'settings', 'signin', 'signout', 'signup', 'userimage', 'search']
```
Arqu: `oauth`, `identity`, `client`, `issues`, `follow`, `unpublish`

Arqu (changes requested):
Can we add these too, once ready, I can merge once I settle things in CI

server/index.js
ramfox: ahhh yes. Sorry I don't want you to this I ignored you when you linked to the backend routes, I must have misunderstood. I didn't think that the react app would ever request these routes. added them!

Arqu (approved):
🚀

PR #451, opened 2021-10-21
Cw/homepage

chriswhong:
- Redesigns `Splash.tsx` with our new landing page design
- Adds custom tailwind css configuration for font sizes based on our style guide

PR #452, opened 2021-10-21
Cw/text size tweaks

chriswhong:
Merge #451 first

This PR includes many updates to tailwind size classes.  We added custom tailwind configuration in #451 to match text sizes in the app with our style guide.  As a result, many changes were needed throughout the app to apply the correct sizes.

Also includes some style updates to the collection view to more closely match the new mockups.


boandriy:
Hey @chriswhong , one thing I noticed when playing around, the dry run button text is wrapping, not sure if we want that. Also during dry run the icon next to the cancel text (dry run text on button turns to cancel during dry run) is much smaller.
image
image

boandriy (approved):
LGTM

chriswhong:
Great catch, this was due to a hard-coded width for this button.  I think it was there to prevent the shift in content when the text changes from "Dry Run" to "Cancel".  I removed it for now.

PR #453, opened 2021-10-21
refactor(collection): update collection based on incoming events

### LINKED ISSUE #285: User can see information about the latest run in collection
The latest run for an automated dataset should appear in the "latest run" column in the Collection

ramfox:
This PR allows the collection view to respond to incoming events that have to do with transform runs, so that we can display to the user that the particular.

This essentially "copies" the behavior of the backend collection. Normally I would not be up for duplicating code like this, but the way we have framed the collection on the backend actually folds in rather nicely to the redux reducer pattern. It prevents unnecessary fetching, and is scoped to only three events: `ETTransformStart`, `ETLogbookWriteRun`, & `ETLogbookWriteCommit`. It would be simple to expand these cases *if we need to* at some point in the future, if we add more functionality to the collection view and would like to prevent fetching of data that already comes across over the websocket connection.

This PR also adds the `runStart` field to `VersionInfo` that we can use along with the `runDuration` to calculate the `runEnd` time.

closes #285 

depends on qri-io/qri#1945

### DIFF
--- a/src/utils/runEndTime.ts
+++ b/src/utils/runEndTime.ts
@@ -1,0 +1,4 @@
+export function runEndTime(runStart: string, runDuration: number): Date {
+  const runStartTime = new Date(runStart)
+  return new Date(runStartTime.getTime() + (runDuration / 1000000))
+}
--- a/src/features/collection/CollectionTable.tsx
+++ b/src/features/collection/CollectionTable.tsx
@@ -17,6 +17,7 @@ import RunStatusBadge from '../run/RunStatusBadge'
 import { VersionInfo } from '../../qri/versionInfo'
 import ManualTriggerButton from '../workflow/ManualTriggerButton'
 import DatasetInfoItem from '../dataset/DatasetInfoItem'
+import { runEndTime } from '../../utils/runEndTime'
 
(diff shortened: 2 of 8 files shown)


src/utils/runEndTime.ts
```diff
@@ -0,0 +1,4 @@
+export function runEndTime(runStart: string, runDuration: number): Date {
+  const runStartTime = new Date(runStart)
+  return new Date(runStartTime.getTime() + (runDuration / 1000000))
+}
```
ramfox, line 4: I know it seems silly to not just have pulled in `RunEnd` rather than `RunStart`, but I figure we will need both at some point and it feels more intuitive to me to add the duration rather than subtract it 🤷‍♀️

chriswhong (approved):
I agree the UX is clumsy when we have things jumping around in the list.  I think this is fine to merge for now, and we'll need to revisit how to effectively meet users expectations when it comes to live changes in a list view.  

We might be able to move the sorting logic to the component (instead of the reducer) so we have more control over when it is applied.  It seems reasonable to me to keep rows in whatever order they were in on page load or the last time the user sorted the list, even if they update.

PR #454, opened 2021-10-21
feat: On Run log, run now button now adds an event to the table

### LINKED ISSUE #418: Run Now instant feedback in Run Log
When the user clicks Run Now in the run log, they may see a toast, but there is no new row in the run log.

I expect a new row to appear at the top of the list, showing that the run is  "Running".  Right now nothing shows up unless the user refreshes the browser (causing new calls to the backend)

boandriy:
On Run log, run now button now adds an event to the table with status running. Added icon next to status in run log table.

closes #418


chriswhong (changes requested):
This looks great when status === 'running' lasts a few seconds.  However, when testing it with a run that happens quickly, the classnames update so fast that the animation is lost.  I don't know what technique we can use make the classnames persist through the running state, but I think it's worth another look.  

Here is some code that will run almost instantly you can test on:

```
load("time.star", "time")
load("dataframe.star", "dataframe")

# append the current time
current_time = time.now()

# get the previous version of this dataset
workingDataset = dataset.latest()

the_list = list(workingDataset.body['time'])
the_list = sorted(the_list + [str(current_time)], reverse=True)
print(the_list)

workingDataset.body = the_list



# finally, commit the changes
# the last step of every transform is always `dataset.commit(Dataset)`
dataset.commit(workingDataset)
```

boandriy:
Hey, @chriswhong , pushed an update that delays a bit to make sure that animation is finished.

chriswhong (approved):
I still think the experience is a bit jumpy with a quick-running workflow, but it's ok for now.  One small point: the row height is no longer 68.  LGTM once we make the row height match.

src/features/activityFeed/DatasetActivityFeed.tsx
```diff
@@ -37,13 +39,39 @@ const DatasetActivityFeed: React.FC<DatasetActivityFeedProps> = ({
     dispatch(runNow(qriRef))
   }
 
+  useEffect(() => {
+    if (latestRun.status === 'running') {
+      const runningLog:LogItem= {
```
b5: missing space

src/features/activityFeed/DatasetActivityFeed.tsx
```diff
@@ -37,13 +39,39 @@ const DatasetActivityFeed: React.FC<DatasetActivityFeedProps> = ({
...
+        timestamp: new Date().toString(),
+        runStatus: "running",
+        title: '--',
+        runID: latestRun.id,
+      } as LogItem
```
b5: instead of coercing to a type, either use or create a `NewLogItem` constructor function whever the `LogItem` type is defined. Even if it's exactly the same, it allows setting default values for types & keeps type coercion creep out of the codebase

b5 (commented):
Two small nits from a technical perspective. I haven't reviewed this for functionality

b5 (approved):
Looks good!

PR #463, opened 2021-10-22
fix: added logic that removes ds from collection on remove

### LINKED ISSUE #462: Dataset delete doesn't immediately update the collection
After you remove a dataset - you can still see the deleted dataset and the dataset count will not change until you restart the page

boandriy:
added logic that removes ds from collection on remove

fixes #462

### DIFF
--- a/src/features/collection/CollectionTable.tsx
+++ b/src/features/collection/CollectionTable.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, {useState} from 'react'
 import { useDispatch } from 'react-redux'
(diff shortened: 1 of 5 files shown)


chriswhong (changes requested):
Maybe I am missing something, but I ran this locally and removing from collection view didn't update the list (the deleted dataset was still present in the list)

boandriy:
@chriswhong , added animation for remove

chriswhong (approved):
The animation looks great! All good from the product side.

ramfox (approved):
LGTM

PR #467, opened 2021-10-22
feat: add custom fathom events

chriswhong:
Adds several fathom events which we can use to evaluate event-tracking

### DIFF
--- a/src/chrome/ButtonGroup.tsx
+++ b/src/chrome/ButtonGroup.tsx
@@ -3,6 +3,7 @@ import { Link } from 'react-router-dom'
 import classNames from 'classnames'
 
 import Icon from './Icon'
+import { trackGoal } from '../features/analytics/analytics'
 
 export interface ButtonGroupProps {
   items: ButtonGroupItem[]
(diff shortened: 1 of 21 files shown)


src/chrome/ButtonGroup.tsx
```diff
@@ -23,9 +23,20 @@ const ButtonGroup: React.FC<ButtonGroupProps> = ({
...
+        <Link
+          className={classNames('my-2 mx-4 font-medium text-black transition-100 transition-all hover:text-qripink', {
+            'text-qripink': selected
+          })}
+          key={i}
+          to={link}
+          onClick={text === 'My Datasets'
+            ? () => {
+              // general-click-my-datasets event
+              window.fathom.trackGoal('LKV9RUBF', 0)
```
b5: we should wrap the call to `fathom.trackGoal` in an abstraction to avoid calling to `window` all over our app. Create a new file `src/features/analytics.ts` & export a wrapper function:

```ts
export function trackGoal(id: string, count number): void {
  window?fathom?trackGoal(id, count)
}
```

The signature of the function should match that of `fathom.trackGoal`. This way if we flip analytics, or try to run this app in an environment that doesn't have the `window` global, it'll be a change in one place. 

Once defined, refactor all calls to `fathom.trackGoal` to use this func instead

src/features/footer/SplashFooter.tsx
```diff
@@ -5,6 +5,12 @@ import Icon from '../../chrome/Icon'
 import QriLogo from '../../chrome/QriLogo'
 
 const SplashFooter = () => {
+
+  const fireFathomEvent = () => {
```
b5: rename this to `fireAnalyticsEvent`

b5 (approved):

PR #469, opened 2021-10-25
fix: resetting collection state on user log out

### LINKED ISSUE #465: Collection is not updated when switching users
- Login and go to collection view
- Log out
- Log in as a different user
- The previous user's collection view persists

We should make sure that the collection is cleared on log out, and re-populated on log in.

boandriy:
resetting collection state on user log out

fixes #465


ramfox (approved):
LGTM

PR #470, opened 2021-10-25
fix: disabling dry run button for users that are not logged in

### LINKED ISSUE #441: refactor(workflow): only logged-in users should be able to "dry run" a workflow
Like the `Deploy` button, a user that is not logged in should be prompted to login or create and account before being allowed to dry run the transform.

boandriy:
disabling dry run button for users that are not logged in

fixes #441 

@chriswhong , the button has no disabled styles, I have changed the tooltip for the button to 'Please log in before running script', Let me know if I should add the styles to disabled button.

image


chriswhong:
Yes, we should make sure each button `type` has a `disabled` style... you should only need to pass in the `disabled` prop.

boandriy:
@chriswhong , updated button disabled styles

chriswhong (changes requested):
I think we need to clarify the desired behavior here. Dry run should only be an option for the dataset owner. Anonymous users or logged-in users who do not own the dataset should see the disabled dry run button.

The text in the tooltip should say "Only the dataset's owner can dry run". Later we can add "forking" or "cloning" so other users can quickly replicate and try whatever public code they find in workflows.

chriswhong (approved):
LGTM

ramfox (approved):
LGTM

PR #475, opened 2021-10-27
feat(workflow): cancel dry runs and runs

### LINKED ISSUE #68: User can cancel dry run
- By clicking the cancel button that appears in place of "Dry Run"
- By keyboard shortcut (Ctrl-C)

ramfox:
closes #68 
depends on qri-io/qri#1949

Clicking "Cancel" during a "dry run" will cancel the run.

Also fixes bug where user could not "dry run" a new workflow, even when logged in.

Also adjusts `/activity` pagination to use `limit` rather than `pageSize`


b5 (approved):
LGTM

PR #477, opened 2021-10-27
fix: disabled workflow for non owner users

### LINKED ISSUE #368: Workflow on dataset not owned by user should be read-only
Users should have a read-only view on the workflow and/or a eventually a button that requests edit access, or at least a warning modal that they are trying to change the settings of a dataset they don't control. 

Right now users are able to tinker with the workflow settings of another user's dataset, however & fortunately, do not appear to be able to save those changes.

boandriy:
disabled workflow for non owner users

fixes #368


chriswhong (approved):
LGTM

b5 (approved):
LGTM!

PR #478, opened 2021-10-27
feat: add 404 page for not found users and datasets

chriswhong:
Adds a 404 page.

Because any url with one segment will match with `/:username` (with `exact` param in the router) and any url with 2 or more segments will match with `/:username/:datasetname` (no `exact` param), we only need to handle rendering 404s in these two routes.

For the first case, we add `error` to `userProfileState`, and check for it in `UserProfile`.  If the server couldn't find the username, we return `NotFoundPage`.

For the second case, we add `error` to `dsPreviewState` and check for it in `DatasetWrapper`.  Any dataset route will show a 404 if the initial call for a preview comes back with a 404.

image


src/features/userProfile/state/userProfileState.ts
```diff
@@ -60,8 +64,10 @@ export const userProfileReducer = createReducer(initialState, {
     state.profile = action.payload.data
     state.loading = false
   },
-  'API_USERPROFILE_FAILURE': (state: UserProfileState) => {
+  'API_USERPROFILE_FAILURE': (state: UserProfileState, action) => {
+    console.log('here', action)
```
ramfox: stray console.log

src/store/api.ts
```diff
@@ -12,6 +12,18 @@ export const CALL_API = Symbol('CALL_API')
 
 export const API_BASE_URL = process.env.REACT_APP_API_BASE_URL || 'http://localhost:2503'
 
+export interface ApiErr {
+  code: number | undefined
```
ramfox: adjust to have `code` be optional:

`code?: number`

src/store/api.ts
```diff
@@ -12,6 +12,18 @@ export const CALL_API = Symbol('CALL_API')
...
+  message: string
+}
+
+export function NewApiErr (): ApiErr {
+  return {
+    code: undefined,
```
ramfox: when `code` becomes options, you no longer have to declare `code` as `undefined` & can remove this line

ramfox (approved):
This will work for cloud, but does not work locally. Looks like much of our expectations around profiles comes from cloud endpoints and not qri core endpoints. 

I think we should be fine for now though, and I'll write an issue noting the differences & asking for discussion on if anything should be done about it.

A couple of small nits/requested changes, but LGTM

PR #481, opened 2021-10-28
refactor: changed collection design

### LINKED ISSUE #460: Collection View Style Tweaks
Includes a few changes to the layout.  
- Removes the large automation icon that was on the left in favor of a smaller one below the dataset name.
- Repositions the automation icon in DatasetVersionInfo to be on the second line (this will affect its placement in other views too)
- Whatever other nudges and changes are necessary to more closely match the mockups

image

boandriy:
changed collection design

closes #460

### DIFF
--- a/src/chrome/DatasetCommitInfo.tsx
+++ b/src/chrome/DatasetCommitInfo.tsx
@@ -11,76 +11,64 @@ interface DatasetCommitInfoProps {
   dataset: Dataset
   // small yields the same info with smaller text, used in collection, history list and run log
   small?: boolean
-  // isRow indicates whether this is being displayed inline in a row in a list (versus in a card)
-  // and is used to tweak styles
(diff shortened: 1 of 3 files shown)


chriswhong (approved):
I added a commit to this to clean things up a bit.  I think there was some confusion because various mockups for the DatasetCommitInfo component were made at different times and they are showing things in different order.

We want the display of DatasetCommitInfo to be consistent throughout the app, it should always have the commit title on the first line and the supporting info on the second line, all in the same order.

I also removed the `flex` and `preview` props, as there should not be a need to control the width of the title... it is using `truncate` and will show ellipses if it is longer than its container.

I also added some tailwind classes to make sure the download button is always present on dataset previews.

ramfox (approved):
LGTM

PR #483, opened 2021-10-28
fix: new workflow cells are editable and not read only

boandriy:
new workflow cells are editable and not read only


chriswhong (approved):
LGTM

Issue #430, opened 2021-10-15
Need a consistent way to know whether a version was created from a workflow/transform


chriswhong:
![image](https://user-images.githubusercontent.com/1833820/137510593-40dc3246-e835-4372-a6e2-08264981d08f.png)

The list on the left is able to use the existence of `runID` in dataset activity to show the automation icon (this is also used in the collection view and activity list).

However, the `CommitSummaryHeader` is derived from the full `dataset` used to render this version, and that does not include `runID` or anything else we can use to show the same icon.

This is inconsistent.  Wherever we show information about a version, we need to be able to show whether that commit was created by a transform vs manual.

chriswhong:
Update:  We have added an `automated` (boolean) prop to `DatasetCommitInfo` which is used to show the automation icon and indicate the commit was machine-made.

The only places we can't show this because we don't have enough information are:
- Above the tabs in the history view
- On Dataset Previews

In all other places we can use `runID` or just show it all the time (e.g. all commits shown in run log are automated)

ramfox:
We do have a `runID` field on the commit! Just need to surface it on frontend.

PR #486, opened 2021-10-28
Cw/next version preview info

### LINKED ISSUE #393: New Version Preview in Workflow Editor should resemble a version preview from history
It should show the commit message, user icon, username, and timestamp.

image

Right now it looks like this:

image

chriswhong:
Builds on work in #481, merge that one first.

- Uses `DatasetCommitInfo` in Workflow Outline to indicate the Next Version Preview is available
- Adds `automated` prop to `DatasetCommitInfo` since `runID` is used inconsistently.  This can be set to true for all workflow dry runs.
- Audits usage of `DatasetCommitInfo` to ensure we are always properly showing `automated` status where we can.  (there are two places where we can't see #430)

image

Closes #393


ramfox (approved):
LGTM!

PR #487, opened 2021-10-28
Cw/featured datasets

### LINKED ISSUE #457: Manually add homepage datasets
The homepage includes placeholder datasets.  Let's pick some to fill the "popular" column.  We may be able to use a real endpoint for the "recently updated" column, but if not we can  hard code those as well.

chriswhong:
Adds three real datasets to each of the homepage "featured" lists

Closes #457


ramfox (approved):
LGTM!

PR #491, opened 2021-10-28
fix: reset 404 error state on new request

chriswhong:
Fixes bug where once a 404 is encountered, all userprofile or dataset routes will also 404 because the state is never reset.


ramfox (approved):
LGTM

PR #492, opened 2021-10-29
refactor(ci): build prod image

Arqu:
Not ideal due to built time/cycles but we need to have different params at build...


b5 (approved):
works for now

PR #496, opened 2021-10-29
refactor: split up history and runs api calls

chriswhong:
Splits up the api calls for history and run log.  Before they were sharing calls to `/activity`, now they each do their own and store the results in their own respective state.

Also adds `loading` and `error` state to both for consistency, but these aren't yet used in the UI.

### DIFF
--- a/src/features/activityFeed/state/activityFeedState.ts
+++ b/src/features/activityFeed/state/activityFeedState.ts
@@ -3,6 +3,8 @@ import { createReducer } from '@reduxjs/toolkit'
 import { RootState } from '../../../store/store';
 import { QriRef, humanRef, refStringFromQriRef } from '../../../qri/ref';
 import { LogItem } from '../../../qri/log';
+import { ApiErr, NewApiErr } from '../../../store/api'
+
 
 export function newDatasetLogsSelector(qriRef: QriRef): (state: RootState) => LogItem[] {
   return (state: RootState) => {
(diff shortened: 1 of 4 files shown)


src/features/activityFeed/state/activityFeedState.ts
```diff
@@ -19,15 +21,29 @@ export function selectLogCount(qriRef: QriRef): (state: RootState) => number {
...
 
 export const activityFeedReducer = createReducer(initialState, {
-  'API_DATASET_ACTIVITY_SUCCESS': (state, action) => {
+  'API_DATASET_ACTIVITY_RUNS_REQUEST': (state, action) => {
+    state.loading = true
+  },
+
+  'API_DATASET_ACTIVITY_RUNS_SUCCESS': (state, action) => {
+    state.loading = false
     const ls = action.payload.data.filter((d: LogItem)  => d.runStatus) as LogItem[]
```
ramfox: we shouldn't need this filter any longer

src/features/commits/state/commitState.ts
```diff
@@ -25,25 +26,24 @@ export function selectDatasetCommitsLoading(state: RootState): boolean {
...
+  'API_DATASET_ACTIVITY_HISTORY_FAILURE': (state, action) => {
     state.loading = false
-    // TODO (b5): capture error from API response
-    state.loadError = 'failed to load commits'
+    state.error = action.payload.err
   },
-  'API_DATASET_ACTIVITY_SUCCESS': (state, action) => {
+  'API_DATASET_ACTIVITY_HISTORY_SUCCESS': (state, action) => {
     const ls = action.payload.data as LogItem[]
     state.commits[refStringFromQriRef(action.qriRef)] = ls.filter(d => d.path)
```
ramfox: we shouldn't need this filter any longer

ramfox (approved):
LGTM

there are two spots, once inside each edited action, where we filter on the resulting version info list. we no longer need to do that, now that filtering is happening on the backend.

PR #502, opened 2021-10-30
fix workflow preview loading state

### LINKED ISSUE #493: Dry run body preview loading state runs forever

chriswhong:
- fixes a perpetual loading state bug that was introduced after the history view loading state was merged.

The workflow "next version preview" Data tab would always show its loading state.  This happened because the logic that it uses is based on a different part of the state tree (`dataset`) which has nothing to do with the current workflow and loading is always `true`

This introduces a prop `showLoadingState` to `TabbedComponentViewer` which ensures the Data tab will never show loading state.  

Closes #493 

Also fixes a useEffect endless loop which was happening in `WorkflowCell`


boandriy (approved):
LGTM!

PR #504, opened 2021-11-01
fix: not disabling the deploy button on type

### LINKED ISSUE #500: Deploy Modal - Dataset name validation causes deploy button to flash with each letter typed
When typing, the validation seems to be setting the button to gray and then back to pink with each letter the user types for the dataset name.

This should be debounced while we are at it.

boandriy:
not disabling the deploy button on type

closes #500


ramfox (approved):
LGTM

PR #509, opened 2021-11-01
fix(activity): show `runEnd` time rather than `commitTime` in the activity log

ramfox:
Also, adds logic to see if a timestamp is actually the "zero" value before displaying it. If it is the "zero" value (or is somehow invalid), instead return "--"

### DIFF
--- a/src/utils/runEndTime.ts
+++ b/src/utils/runEndTime.ts
@@ -1,4 +1,5 @@
 export function runEndTime(runStart: string, runDuration: number): Date {
   const runStartTime = new Date(runStart)
+  if (!runDuration) { runDuration = 0 }
   return new Date(runStartTime.getTime() + (runDuration / 1000000))
 }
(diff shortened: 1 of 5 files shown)


src/utils/runEndTime.ts
```diff
@@ -1,4 +1,6 @@
 export function runEndTime(runStart: string, runDuration: number): Date {
   const runStartTime = new Date(runStart)
-  return new Date(runStartTime.getTime() + (runDuration / 1000000))
+  if (!runDuration) runDuration = 0
```
chriswhong: Add curly brackets even for one line if statements

src/utils/runEndTime.ts
```diff
@@ -1,4 +1,6 @@
...
+  const d =  new Date(runStartTime.getTime() + (runDuration / 1000000))
+  return d
```
chriswhong: Maybe just `return new Date(runStartTime.getTime() + (runDuration / 1000000))`

chriswhong (approved):
2 nits, then LGTM

src/utils/runEndTime.ts
ramfox: ahh yes, ty. Hold over from when I was testing & needed log the output

PR #517, opened 2021-11-02
fix: remove execute from monaco editor context menu

### LINKED ISSUE #499: Code Block context menu
There's a z-index issue here, but there is also an option to "execute block" which isn't a thing.  Let's figure out how to remove this option from the menu and fix the z-index problem.

image

boandriy:
remove execute from monaco editor context menu

fixes #499


ramfox (approved):
LGTM

This only fixes the "remove execute and block option" part of this issue. This doesn't address the zindex problem.

chriswhong (approved):
Tested this locally, looks great.

PR #518, opened 2021-11-02
fix: disabling the delete cell button if the call is the last one

### LINKED ISSUE #484: [workflow]: no way to add a new cell if user removes all cells.
Once the last cell is removed, there is no way to add a new one.

I think "Remove" should simply be disabled if the number of cells === 1.  We should not let the user get into a state where they have 0 cells.

boandriy:
disabling the delete cell button if the call is the last one

Fixes #484


ramfox (approved):
LGTM

PR #520, opened 2021-11-02
fix(activity): only the dataset owner can "Run Now" from the activity feed

### LINKED ISSUE #515: Run Now button in Run Log should be disabled for non-owner
The Run log is public, but the Run Now button should be disabled when the user does not own the dataset.  It should show a similar message "Only the dataset owner can trigger runs" or whatever language we used for the dry run button in the workflow editor.

ramfox:
closes #515


boandriy (approved):
LGTM!

chriswhong:
Can we make the tooltip say "Only the dataset owner can trigger a run"?  I don't think "script" is the word to use here.

PR #528, opened 2021-11-02
chore: add standard, react, typescript eslint rules

### LINKED ISSUE #339: need to add `ts-standard` as our linter
We no longer have a linter standardizing the way we format our typescript code.

We have previously used `ts-standard`: https://standardjs.com/#typescript

chriswhong:
This is a monster PR which configures `eslintConfig` to use `standard-with-typescript` and `plugin:react/recommended`.  These are the same eslint packages we used in Qri Desktop.

This of course introduces a whole new set of linting and typescript errors which prevent build successful development and production builds, so this PR squashes all of them by either:
- adding/editing types, checking for types, making types optional, etc to silence the errors
- using `eslint-disable-next-line` in some places to just ignore certain linting errors
- using `@ts-expect-error` in some places to to just ignore certain ts errors
- using `@ts-nocheck` on some files, many of these are unused components that were brought over from Qri Desktop but were never used, so they have lots of dependencies that don't exist or outdated types, etc.

I realize this is a chaotic PR, but I think this will be necessary to get us to a clean state with no TS errors.  Moving forward we should take some extra time to make sure every file we touch has appropriate types and no superfluous ignore comments. 

Closes #339


boandriy (approved):
LGTM

ramfox (approved):
this is all of my dreams come true

we should do a follow up (not now) that hunts for some of these ts/es lint rules and sees if they are worth keeping or following

LGTM

PR #541, opened 2021-11-04
feat: added option to edit ds title

### LINKED ISSUE #263: User can edit dataset title
This feature allows the user to edit the `meta.title` of a dataset, which will result in a commit and a new version of the dataset.  

Frame 627

Flow:
- User clicks the title of a dataset they own
- The editor opens "in place", so the text input is directly on top of the original text, and is pre-populated with the current title.  The text should be pre-selected so the user can begin editing immediately.
- User can close the editor by clicking outside of it or pressing escape (we have this functionality in `Modal`, we may want to make this editor a `type` of Modal but need to add a way to position it precisely)
- When the user clicks commit, fire an action
- On success or failure, close the UI
- TODO: display loading state, display success/error message.  

Action:
- add a new action `commitDatasetTitle(qriRef: QriRef, title: string)`
- it will send a `POST` to `/ds/save` with a body like:

```
{
    "ref": `${qriRef.username}/${qriRef.name}`,
    "title": "I added this title via ds/save"
}
```
Reducer:
- handle `API_COMMIT_DATASET_TITLE_SUCCESS`.  If the commit was successful, we need to figure out how to update the dataset in state since we are now working with a new HEAD version.  The actions used for this happen in `DatasetWrapper`, so we probably need to add some state that invalidates the current HEAD and add it to the `useEffect` in `DatasetWrapper`, causing it to re-fetch.

boandriy:
added option to edit ds title

Closes #263


chriswhong (changes requested):
- [x] The modal is not positioned over the text.  The goal is for this to feel like an "edit in place" UI, as if clicking the title will expose an input in the same place.  The only difference is that we are adding a button and a box around it.  This will take some work to figure out how to position it properly.
- [x] The commit title is not editable.  I think the user needs to be able to edit this in case they need to explain why they are making the change.
- [x] Change the text to say "This title is part of the dataset's `meta` component.  Editing it will create a new commit."  (we may need a component for making "meta" appear like a code block.  See our MDX definition in website to get matching tailwind styles.

chriswhong:
- [x] The editing icon is not vertically centered
- [x] Let's only show the editing icon when hovering over the title
image
- [x] Input ring is blue, let's make it match the others in the app (gray, pink when it has focus). We already have a good `TextInput` component that has all of the styles already in place.  I would suggest seeing if we can use this component and pass in new props or styles to make it work for this application
- [x] I don't think the modal needs to be this wide.  75% of the current width would probably be ok.

chriswhong (changes requested):
This is working great.  I think the only bit that caught my eye is that there is no loading state while the request is being made, and it looks like the title is cleared before the new one is applied (so it looks blank for a brief second)

Can we add a loading spinner while the modal is open?  When the user clicks commit, they should see the spinner and the modal should disappear only when the change is successful.  We will use this pattern again with all manual commits, so it's probably best to get it worked out now.

Same thing for error handling, if the change is not successful, we should show the error in the modal, and the user can either cancel or try again.

chriswhong:
Pushed a commit with the following:
- rebase to origin/master
- Re-uses TextInput with new `size` prop for commit title
- Adds a set width to the button
- Moves the error to its own div at the bottom of the modal (because it is not related specifically to either input)
- Disables the button until the title is changed
- Don't show edit icon if title is not `editable`

src/chrome/forms/TextInput.tsx
```diff
@@ -48,15 +52,18 @@ const TextInput: React.FC<TextInputProps> = ({
...
         {label && <InputLabel
           label={label}
           tooltip={labelTooltip}
           tooltipFor={name}
         />}
         <input
-          className="focus:ring-transparent focus:border-qripink-600 block w-full px-2 text-sm border-qrigray-300 rounded-md placeholder-qrigray-400 rounded-lg"
+          className={`focus:ring-transparent focus:border-qripink-600 block w-full px-2 text-sm border-qrigray-300 rounded-md placeholder-qrigray-400 rounded-lg ${inputClassName}`}
+          style={{
+            height: (size === 'sm') ? 24 : ''
```
ramfox, line 65: we have size options sm md & lg, but then only have a value for sm? I'm just a bit confused by this. can you explain to me what the differences between each size should be?

src/features/dataset/state/datasetActions.ts
```diff
@@ -111,6 +143,30 @@ function fetchBody (ref: QriRef, page: number, pageSize: number): ApiAction {
...
+export function commitDatasetTitle (qriRef: QriRef, title: string, commitTitle: string): ApiActionThunk {
+  return async (dispatch) => {
+    const action = {
+      type: 'title',
+      [CALL_API]: {
+        endpoint: 'ds/save',
+        method: 'POST',
+        body: {
+          ref: `${qriRef.username}/${qriRef.name}`,
+          message: commitTitle,
```
ramfox: Why do we have the message and title as the same thing? LMK if this was requested by product, because it is possible to only set the title & not also be forced to send the message as well.

src/features/dataset/state/datasetActions.ts
```diff
@@ -43,6 +44,37 @@ function getExistingHeader (collection: Record<string, VersionInfo>, ref: QriRef
...
+      type: 'rename',
+      [CALL_API]: {
+        endpoint: 'ds/rename',
+        method: 'POST',
+        body: {
+          current: refStringFromQriRef(humanRef(current)),
+          next: refStringFromQriRef(humanRef(next))
+        },
+        // TODO(b5): this return value is a versionInfo
+        map: mapDataset
```
ramfox: replace this with `mapVersionInfo` & remove the comment :)

src/features/dataset/DatasetHeader.tsx
```diff
@@ -45,14 +65,42 @@ const DatasetHeader: React.FC<DatasetHeaderProps> = ({
...
+          dispatch(loadDatasetCommits(ref))
+        }
+      })
+  }
+
+  const onTitleClick = () => {
+    if (!isNew) {
+      const left = location.pathname === `/${header.username}/${header.name}` ? '-206px' : '153px'
+      const top = location.pathname === `/${header.username}/${header.name}` ? '67px' : '65px'
+      const position = location.pathname === `/${header.username}/${header.name}` ? 'relative' : 'absolute'
```
ramfox: so this is an interesting way of solving the location problem, but I'm worried that if we manipulate the page in the future, these numbers will be out of date.

Have you looked into using a callback to measure a DOM element: https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node

ramfox (changes requested):
looking great!

Some questions and only suggestion for refactoring!

src/features/dataset/state/datasetActions.ts
chriswhong: We should not be sending the `message` as part of this request, `title` only.

src/chrome/forms/TextInput.tsx
chriswhong: I did this change, I needed to add  `sm` for a shorter text input, and implied that the default input would be `md`.  `lg` doesn't exist but I left it in there in case we need it.  Can take it out.

src/features/dataset/DatasetHeader.tsx
chriswhong: +1

src/features/dataset/DatasetHeader.tsx
```diff
@@ -31,9 +47,21 @@ const DatasetHeader: React.FC<DatasetHeaderProps> = ({
...
+  const dataset = useSelector(selectDataset)
+  const headerLoading = useSelector(selectIsHeaderLoading)
+  const datasetLoading = useSelector(selectIsDatasetLoading)
   const qriRef = qriRefFromVersionInfo(header)
+  const [ titleChangePosition, setTitleModalPosition ] = useState({ top: '0', left: '0' })
+
+  const measuredRef = useCallback(node => {
+    if (node !== null) {
+      const left: string = (node.getBoundingClientRect().left - 25).toString() + 'px'
+      const top: string = (node.getBoundingClientRect().top - 56).toString() + 'px'
```
ramfox: very cool!

Can I ask where the `25` & `56` come from? 

I'm asking because this works perfectly on chrome, but on safari and firefox the text jumps down a few pixels.

boandriy: I think it comes from paddings/ borders of the popup. @ramfox

ramfox: I let me re-phrase: did these numbers come from experimentation or do they come from knowing the pixels of the line height or something. Because, again, they work on chrome, but are different for safari and firefox

boandriy: I was experimenting

boandriy: I see the problem. Do you have any suggestions?

boandriy: This is how it calculates it on Chrome: 
left: 280.045
top: 118.864
This is on firefox: 
left: 178.5
top: 119.1

ramfox: thanks for doing the further research.

I've unfortunately realized we have a few more problems using this approach. Basically, this won't recalculate if you toggle the sidebar nav over, or if you make the browser window small & the smaller scrolling header folds down:
Screen Shot 2021-11-10 at 9 04 57 PM
(side question: is that a bug, should the two headers be open at the same time?)

But, since we know that the two potential changes to the position are when the sidebar goes in and out or the `DatasetMiniHeader` goes in or out, we can use the state and selectors to know whether they have changed from in to out or back, and we can recalculate then.

We already have the sidebar UI in the `App` state tree, so we can just `useSelector(selectNavExpanded)` to get `isNavExpanded`.

We can add a field `isMiniHeaderInView` (or some other name), to the `Dataset` state & add a `setIsMiniHeaderInView` action, that sets the field in the `Dataset` state, and create a `selectIsMiniHeaderInView` selector. 

Then, like we do for the `isNavExpanded`, we use `useSelector(selectIsMiniHeaderInView)` to get `isMiniHeaderInView` in `DatasetHeader`

Then, we just add `isNavExpanded` & `isMiniHeaderInView`  to the dependency array of the `useCallback` & that's one problem solved.

This doesn't fix the pixel differences between safari/firefox and chrome. I got side tracked when I discovered the above problem, lol.

chriswhong (approved):
LGTM 🚀

PR #544, opened 2021-11-04
feat: responsive splash page

chriswhong:
- Makes the splash page responsive
- We had a global min-width and a fixed height throughout the app.  To make the landing page responsive, it needs to be a long-scroller, so some refactoring of the rest of the app's layout was necessary.  This work introduces `WebAppLayout`, which wraps its children with a div with the correct height and min-width.  This is used for now on all dataset routes, so things like search results and user profiles can exist without it and look better on mobile.


src/features/splash/Splash.tsx
```diff
@@ -53,11 +64,11 @@ const Splash: React.FC<{}> = () => {
                 </div>
               </div>
             </div>
+            <div className='my-16 md:hidden'>&nbsp;</div>
           </div>
-          <div className = 'flex flex-col flex-shrink-0' style={{ width: 510 }}>
-            <NavBar minimal transparent noLogo />
-            <div className='font-bold text-3xl mb-6 px-9'>Discover Datasets</div>
-            <div className='mb-4 flex items-center px-9'>
+          <div className = 'flex flex-col flex-shrink-0 md:mt-24 md:w-5/12 lg:w-1/2' style={{ maxWidth: 754 }}>
```
boandriy, line 69: Hey @chriswhong , I see issues with scrolling the ds container. I guess this is because of the absolute image that is really big. Because if you put here classname of relative and add a z-50 or whatever is grater than blob on the left
`<div className = 'relative z-50 flex flex-col flex-shrink-0 md:mt-24 md:w-5/12 lg:w-1/2' style={{ maxWidth: 754 }}>` 
then the content is scrollable.

boandriy (approved):
LGTM! But would be great if that scrolling thing would be fixed ^

PR #559, opened 2021-11-08
fix: minor styles fixes

### LINKED ISSUE #534: PunchList
For small issues that don't warrant their own issue:

- [ ] [splash]: investigate missing background blobs on mobile safari
- [x] [footer]: Remove "Subscribe" from footer since there is no subscribe input
- [x] [toast]: The sub text on the toast when it’s successful, failed or running it is navy and should be black
- [x] [preview]: If there are no keywords, show "no keywords" in the metadata pane
- [x] [search]: remove `placeholder-opacity-50` in SearchBox

image
- [x] Vertically center text in pagination buttons

image

- [x] Disable workflow in datasetnav for datasets that the user does not own.  Keep it in place with its tooltip, just disable it.

boandriy:
minor styles fixes

Skipped the [splash]: investigate missing background blobs on mobile safari - task. Spend some time investigating. Turns out that <filter> tag in svg is not supported on safari browser. 

fixes #534


src/chrome/PageControl.tsx
```diff
@@ -59,7 +59,7 @@ const PageControl: React.FC<PageControlProps> = ({
             pageCount={pageCount}
             pageRangeDisplayed={5}
             marginPagesDisplayed={2}
-            pageClassName='bg-gray-200 px-2.5 py-2 font-semibold transition rounded-sm m-1 leading-3 hover:bg-white hover:text-dangerred'
+            pageClassName='pagination_controls bg-gray-200 px-2.5 py-2 font-semibold transition rounded-sm m-1 leading-3 hover:bg-white hover:text-dangerred'
```
chriswhong, line 62: I don't think this is the right approach.  To vertically center the text, just make sure the height of the container is equal to the line height of the text.  (add an inline `style` attribute setting the `lineheight` and `height` to the same value

src/index.css
```diff
@@ -1158,3 +1158,7 @@ code {
 .min-height-0-important {
   min-height: 0 !important;
 }
+
+.pagination_controls{
```
chriswhong, line 1162: See above

chriswhong (changes requested):

src/chrome/PageControl.tsx
boandriy: The ReactPaginate component doesn't accept inline styles. This was my first approach. @chriswhong

chriswhong (approved):

ramfox (approved):
LGTM

PR #561, opened 2021-11-08
fix: made ds preview tabs links take full width

### LINKED ISSUE #546: [workflow] Clickable tab zone
The only thing clickable on the tabs is the text and the icon, It should be the whole tab
Screenshot 2021-11-04 at 17 49 51

boandriy:
made ds preview tabs links take full width

fixes #546


ramfox (approved):
LGTM

PR #564, opened 2021-11-08
fix(*): add `ExecutionMode` and show user photo in `UserProfile` component

### LINKED ISSUE #535: [auth]: User profile images missing
Users should have been assigned a profile image.  We have no UI to change them, but the one we have assigned should show up in the app.  

image
image

ramfox:
This pr takes a few basic steps to allow us to split behavior between a local, desktop, and cloud version of the frontend app.

It adds an `ExecutionMode` enum with options `LOCAL`, `DESKTOP`, and `CLOUD`

To allow local to fetch identity, dataset_summary, and follow information correctly, add a hack that, if we are NOT in `CLOUD` execution mode, will fetch those endpoints from the `API_BASE_CLOUD_URL`

This does not solve all of our user photo problems. We are still only showing hard coded values in the commit (using <UsernameWithIcon>). Solving this will require some refactoring to the state tree, so I want to do it in a follow up.

closes #535

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,7 @@ ARG app_port
 ENV REACT_APP_API_BASE_URL=$api_base_url
 ENV REACT_APP_WS_BASE_URL=$ws_base_url
 ENV PORT=$app_port
+ENV APP_EXEC_MODE="CLOUD"
 
 # create app directory
 WORKDIR /app
--- a/src/execution/execution.ts
+++ b/src/execution/execution.ts
@@ -1,0 +1,22 @@
+export enum ExecutionMode {
+  // LOCAL represents running the app as a browser application running locally
+  LOCAL = "LOCAL",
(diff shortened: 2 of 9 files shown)


Arqu:
Tbh this really smells like we need a feature flags mechanic. Have a common backend endpoint which will return a list of valid features which you can then selectively use on the frontend.

ramfox:
> Tbh this really smells like we need a feature flags mechanic. Have a common backend endpoint which will return a list of valid features which you can then selectively use on the frontend.

We definitely need this and we are currently `ExecutionMode` to bandaid over not having a valid features flag on the backend, but there is another layer here when we add a desktop build on top of a local browser build, the backend won't know if frontend is running as a browser app or electron app (for example). So while we aren't using `ExecutionMode` in this way currently, we will need it regardless if we have a feature flag in the future!

b5 (approved):
LGTM

PR #566, opened 2021-11-09
feat: added hover state to collections links

### LINKED ISSUE #511: Hover state for collection row links
In Collection, clicking in the Name, Last Update, and Last Run column do different things, but the only highlight we have is place is to change the background color of the entire row.

We should indicate to the user that each of these will link to different things by highlighting them independently when hovered.  One idea is to just underline the text in each section on hover, or we could do more.  Need some design input.

boandriy:
added hover state to collections links

closes #511


ramfox (approved):
LGTM!

PR #568, opened 2021-11-09
fix: after run now click running log will appear all the time

### LINKED ISSUE #418: Run Now instant feedback in Run Log

boandriy:
after run now click running log will appear all the time

fixes #418


chriswhong (approved):
Looks great!!!

ramfox (approved):
lovely! LGTM!

PR #570, opened 2021-11-09
feat(users): add `Users` state tree & use profiles to power showing icons

### LINKED ISSUE #535: [auth]: User profile images missing

ramfox:
- creates a state tree section called `Users` that we can use to store multiple profiles.
- because we can store multiple profiles, we can now load profiles for icons on the history and collection pages
- lazy fetches profiles: if we already have it we don't fetch it each time
- adds `DEFAULT_PROFILE_PHOTO_URL` as a stopgap between now and when we can generate icons for users with no photo

### DIFF
--- a/src/chrome/UsernameWithIcon.tsx
+++ b/src/chrome/UsernameWithIcon.tsx
@@ -1,30 +1,42 @@
-import React from 'react'
+import React, { useEffect } from 'react'
 import classNames from 'classnames'
+import { useDispatch, useSelector } from 'react-redux'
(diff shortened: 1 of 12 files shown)


ramfox:
since we want a uniform way of handling loading user icons, not only in the places that exist currently in the app, but as we move forward (eg search items), this is ultimately not how we want to solve loading user icons.

PR #572, opened 2021-11-09
feat: add use hash to assign user icon

chriswhong:
- Adds a util `assignUserIcon` which hashes a `username` and assigns one of our 7 default blob icons. 
- This is then used everywhere a user icon is shown in the app, achieving consistency until we can add custom/user-defined icons

(This eliminates all uses of placeholder user icons throughout the app)


ramfox (approved):
Fab!

LGTM!

PR #574, opened 2021-11-10
feat: added logic to fetch slash datasets

### LINKED ISSUE #488: Display real featured/recent/automated datasets on homepage
This is the old cloud endpoint that powered featured/recent datasets: https://rosebud-api.qri.cloud/dataset_summary/splash

We should:
- update the endpoint to include `popular`, `latest`, `automated` we can define the rules for these as we build it but I think popular is based on download count, latest is self-explanatory, and automated is popular AND automated (assuming this doesn't get us the same list as `popular`

- Consume this endpoint on the homepage.

boandriy:
added logic to fetch slash datasets

closes #488


chriswhong (approved):

ramfox (approved):
LGTM!

Issue #576, opened 2021-11-10
Crash on Dry Run


chriswhong:
Dry run this code:

```py
# load starlark dependencies
load("http.star", "http")
load("encoding/csv.star", "csv")
load("encoding/json.star", "json")
load("re.star", "re")
load("time.star", "time")

# hit the openweathermap API to get data for Brooklyn, NYC
csvDownloadUrl = "https://api.openweathermap.org/data/2.5/weather?q=Brooklyn,ny,usa&APPID=7c350ee6493701b2c9be3e8ed9c5f857&units=imperial"

raw_json = json.decode(http.get(csvDownloadUrl).body())

# set the body
temp = raw_json['main']['temp']
humidity = raw_json['main']['humidity']

ds = dataset.latest()

# append the new data as a row
ds.body.append([[str(time.now()), temp, humidity]])

# set the dataset body
dataset.commit(ds)
```
Error:

```
2.1cdaaffc.chunk.js:2 TypeError: Cannot read properties of undefined (reading 'length')
    at main.97a6ea3c.chunk.js:1
    at la (main.97a6ea3c.chunk.js:1)
    at Va (main.97a6ea3c.chunk.js:1)
    at ra (2.1cdaaffc.chunk.js:2)
    at qs (2.1cdaaffc.chunk.js:2)
    at Du (2.1cdaaffc.chunk.js:2)
    at Su (2.1cdaaffc.chunk.js:2)
    at Ou (2.1cdaaffc.chunk.js:2)
    at vu (2.1cdaaffc.chunk.js:2)
    at 2.1cdaaffc.chunk.js:2
ss @ 2.1cdaaffc.chunk.js:2
n.callback @ 2.1cdaaffc.chunk.js:2
ho @ 2.1cdaaffc.chunk.js:2
ps @ 2.1cdaaffc.chunk.js:2
ju @ 2.1cdaaffc.chunk.js:2
t.unstable_runWithPriority @ 2.1cdaaffc.chunk.js:2
Hr @ 2.1cdaaffc.chunk.js:2
Lu @ 2.1cdaaffc.chunk.js:2
vu @ 2.1cdaaffc.chunk.js:2
(anonymous) @ 2.1cdaaffc.chunk.js:2
t.unstable_runWithPriority @ 2.1cdaaffc.chunk.js:2
Hr @ 2.1cdaaffc.chunk.js:2
Kr @ 2.1cdaaffc.chunk.js:2
[...]

PR #577, opened 2021-11-11
fix(qri): add `runID` field to `Commit` interface

### LINKED ISSUE #430: Need a consistent way to know whether a version was created from a workflow/transform

ramfox:
closes #430

We already have a `runID` field on the `commit` in the backend, we just needed to surface it on the frontend!

Also closes the frontend portion of #576

---

Some notes:

a bunch of these refactors are to get `LogItem` in line with `VersionInfo`, which is the actual data structure that come back from the backend. I think with these changes we can mostly get rid of any references to `LogItem` in favor of using `VersionInfo`, but wanted to make a note of the changes and not do the entire refactor in one sweep, which can be jarring if everyone working on the code base isn't made aware of the change before it happens.

Why should we refactor away from `LogItem`? For long term maintenance, its easier to work with as few data models as possible. Since `VersionInfo` is the actual data structure coming from the backend, and it is exactly the same as the `LogItem`, it will be easier for us in the long term to slim down to using just one, and only create a separate data structure when we absolutely need it.

### DIFF
--- a/src/chrome/DatasetCommitInfo.tsx
+++ b/src/chrome/DatasetCommitInfo.tsx
@@ -5,10 +5,11 @@ import Icon from './Icon'
 import RelativeTimestampWithIcon from './RelativeTimestampWithIcon'
 import UsernameWithIcon from './UsernameWithIcon'
 import commitishFromPath from '../utils/commitishFromPath'
-import { Dataset } from '../qri/dataset'
+import { LogItem } from '../qri/log'
+import { VersionInfo } from '../qri/versionInfo'
 
 interface DatasetCommitInfoProps {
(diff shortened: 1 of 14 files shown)


boandriy (approved):
LGTM

boandriy (approved):

PR #579, opened 2021-11-12
run log inspection

### LINKED ISSUE #265: User can inspect individual run logs
Users need to inspect runs and see a console-style output to see what happened and when.  Each row in the run log can be expanded, revealing a scrollable terminal window.

boandriy:
runlog incpect on row click

Will Close #265

### DIFF
--- a/src/features/activityFeed/state/activityFeedState.ts
+++ b/src/features/activityFeed/state/activityFeedState.ts
@@ -111,7 +111,7 @@ export const activityFeedReducer = createReducer(initialState, {
   'API_DATASET_ACTIVITY_RUNINFO_FAILURE': (state, action) => {
     const runID = action.payload.requestID
     state.runInfo[runID] = {
-      error: action.payload.error
+      error: action.payload.err
     }
   }
 })
--- a/src/features/activityFeed/state/activityFeedState.ts
+++ b/src/features/activityFeed/state/activityFeedState.ts
@@ -3,7 +3,9 @@ import { createReducer } from '@reduxjs/toolkit'
 import { RootState } from '../../../store/store'
 import { QriRef, humanRef, refStringFromQriRef } from '../../../qri/ref'
 import { LogItem } from '../../../qri/log'
+import { Run } from '../../../qri/run'
 import { ApiErr, NewApiErr } from '../../../store/api'
+import formatRunLogTimestamp from '../../../utils/formatRunLogTimestamp'
 
 export function newDatasetLogsSelector (qriRef: QriRef): (state: RootState) => LogItem[] {
   return (state: RootState) => {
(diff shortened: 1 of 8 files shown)


b5:
backend is going to be a blocker on this for at least a week. Can we hide this behind a feature flag, in a similar fashion to https://github.com/qri-io/frontend/blob/3540ca4a6a66df1048ce98a0d55f35d0e783cd9d/src/features/dataset/DatasetRoutes.tsx#L67-L71

chriswhong:
Added commits to improve the expansion UI (added an arrow icon to toggle visibility of run logs) and action/reducer to consume run log data from `/auto/runinfo`.

Each run should now be inspectable in the UI.

TODO:

- [x] error handling - when the api call returns an error, display the error/"unable to retrieve logs" message in the UI
- [x] max height/overflow-y - we need to test this with larger logs to make sure we are showing enough and the outputs are parseable.

**Do not merge until https://github.com/qri-io/qri/issues/1989 is fixed, AND we have removed the `ts-ignore` in `activityFeedState`**

chriswhong (approved):
LGTM!  The height of the logs is fine for now, we can tweak later after this gets used in the wild and we get some feedback.

src/features/activityFeed/state/activityFeedState.ts
```diff
@@ -18,16 +20,63 @@ export function selectLogCount (qriRef: QriRef): (state: RootState) => number {
...
+    stepString += `[step ${i}     ]\n`
+    stepString += `[name       ] '${step.name}'\n`
+    stepString += `[started    ] ${step.startTime}\n`
+    stepString += `[ended      ] ${step.stopTime}\n`
+    stepString += `[status     ] ${step.status}\n`
+
+    if (step.output) {
+      stepString += step.output.reduce((acc, output) => {
+        let outputString = acc
+        // TODO(chriswhong): calls to /auto/runinfo are returning Timestamp and Type keys, this needs to be fixed on the backend
```
ramfox, line 74: good find.

Asmir has a pr to fix this, but I wanted to give you a heads up, I'm advocating we use `timestamp` (rather than `Timestamp` &or `ts`) everywhere we use `events` in the backend codebase, which will require some front end refactoring when it goes live. Ultimately, it means we can use the same `EventLogLine` typescript interface everywhere we interact with events on the frontend, rather than only being able to use it when looking at events coming over the websocket connection.

src/utils/formatRunLogTimestamp.ts
```diff
@@ -0,0 +1,7 @@
+// converts a runlog's Timestamp to a formatted string, used in displaying run logs
+import { fromUnixTime, format } from 'date-fns'
+
+export default function formatRunLogTimestamp (t: number): string {
```
ramfox, line 6: you can import the util function `toMilliFromNano` & use it to convert the timestamp from nanoseconds to milliseconds (which works with the node `Date` function):

`format(new Date(toMilliFromNano(t))), "yyyy-MM-dd'T'HH:mm:ss.SSSxxx")`

ramfox (approved):
I have one requested chance (in `formatRunLogTimestamp`, but then LGTM.

Future refactoring will need to happen once backend aligns its serialization of events.

PR #582, opened 2021-11-15
refactor(routes): add `PrivateRoute` to `DatasetRoutes`

### LINKED ISSUE #536: [auth]: Redirect user on sign out
When the user signs out, if they are on a private route (like collection) they should be redirected to the splash page.

If they are on a public route (dataset preview) they can stay on that route, but we need to make sure the UI changes to reflect that they are logged out (they should no longer be able to edit if they were viewing their own preview)

ramfox:
if a user logs out & they are on a route that is meant to be private, they will be redirected to the route route `/`

closes #536


chriswhong (approved):
LGTM!

PR #591, opened 2021-11-18
wip: added manual dataset creation

boandriy:
added manual dataset creation UI


src/features/manualDatasetCreation/state/manualDatasetCreationActions.ts
```diff
@@ -0,0 +1,72 @@
...
+    meta
+  }
+}
+
+export function saveManualDataset (bodyFile: File): ApiActionThunk {
+  return async (dispatch, getState) => {
+    return dispatch({
+      type: 'saveform',
+      [CALL_API]: {
+        endpoint: 'ds/save/form',
```
ramfox: this endpoint name has changed, it's now `ds/save/upload`

src/features/manualDatasetCreation/state/manualDatasetCreationActions.ts
```diff
@@ -0,0 +1,72 @@
...
+        method: 'POST',
+        form: {
+          'ref': 'chriswhong-demo/my-foo',
+          'bodyFile': bodyFile
```
ramfox: this form key has changed, it's now expecting 'body' rather than 'bodyFile'

Just for your knowledge if we want to expand in the future, we can also now add files for 'readme', 'transform', 'viz', and a dataset.json file using the key 'file', sent to this same endpoint.

src/features/manualDatasetCreation/state/manualDatasetCreationActions.ts
```diff
@@ -0,0 +1,72 @@
...
+  return {
+    type: SET_MANUAL_DATASET_CREATION_META,
+    meta
+  }
+}
+
+export function saveManualDataset (bodyFile: File): ApiActionThunk {
+  return async (dispatch, getState) => {
+    return dispatch({
+      type: 'saveform',
```
ramfox: not necessary to change it, but just so you know if you want to match the backend, on the backend we now refer to this as 'saveupload', rather than 'saveform'

chriswhong:
I pushed a commit with some major refactoring:
- Splits `DatasetHeader` into two components: `DatasetHeader`, which contains all of the selectors and actions used when rendering the Header for the active dataset and `DatasetHeaderRender` which is a simplified rendering component.
- `DatasetHeaderRender` has some improvements to the dataset name and title.  Now they are *always* text inputs that change their styles when editable/focused.  This prevents the shift we were seeing when editing these values.  Also added styles to make them match the mockups/color scheme.
- `DatasetHeaderRender` is then used by `ManualDatasetCreation` to allow for editing of the name and title.

When the user selects a file to upload:
- we take the filename and use it to replace both the dataset name and title, but only if they have not been changed from the default values.  If the user has already changed one of these fields, it will not be changed when the file is added.
- we also update the commit title to include the file name, which is a nice way to add some traceability

**Note** For now I have removed the ability to edit a dataset's title from the header on a dataset the user owns.  I have kept all of the UI and logic for this in `DatasetNameCommitModal` which is unused, but we can bring this back if we need it.  I think what we will probably want here is to just bring the user to the dataset editor UI if they want to edit the meta.title.

chriswhong:
To get this PR across the finish line:
- [ ] Add the "close" (X) `IconButton` as a child of `DatasetHeaderRender` to allow the user to cancel creating a new dataset.  This should just navigate to the previous route (it's essentially a back button)
- [ ] Add the modal that appears on commit.  See the video posted by Tefy in #product. 
  - This modal should appear when the user clicks commit
  - This modal should not have any input fields in it, as we have already updated the name and title based on the filename
  - This modal should show confetti once the request comes back successful and the dataset has been created.
  - The final screen should say "Dataset Created" and give the user two buttons: "Add Metadata" and "Show Dataset Preview"

src/features/dsComponents/ComponentItem.tsx
```diff
@@ -19,6 +19,7 @@ export interface ComponentItemProps {
   tooltip?: string
   // for showing a gray border around the selected tab to contrast with white background
   border?: boolean
+  manualCreation?: boolean
```
ramfox, line 22: we pass in `manualCreate` (and `color`), and then don't use them. Maybe remnants from a previous refactor? They should be removed.

src/features/manualDatasetCreation/CompletionCheckmark.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react"
+
+interface CompletionChenckmarkProps {
```
ramfox: probs should read `CompletionCheckmarkProps` (currently "Chenckmark")

src/features/dsComponents/ComponentList.tsx
```diff
@@ -85,19 +87,34 @@ const ComponentList: React.FC<ComponentListProps> = ({
...
               />
           )
-        }
-        return (
-          <ComponentItem
+        } else if (manualCreation) {
+          return (
+            <ComponentItem
+              disabled={name !== 'body'}
+              manualCreation
```
ramfox: same as above

src/features/dsComponents/ComponentList.tsx
```diff
@@ -85,19 +87,34 @@ const ComponentList: React.FC<ComponentListProps> = ({
...
+              displayName={displayName}
+              icon={icon}
+              selected={selectedComponent === name}
+              border={border}
+            />
+          )
+        } else {
+          return (
+            <ComponentItem
+              manualCreation
```
ramfox: same as above

src/chrome/EditableLabel.tsx
```diff
@@ -1,26 +1,32 @@
...
-  value: string
+  value?: string
   size?: 'sm' | 'md' | 'lg' | 'xl'
   readOnly?: boolean
   validator?: (name: string) => ValidationError
-  onChange: (name: string, value: string) => void
+  onChange?: (name: string, value: string) => void
+  placeholder?: string
+  textClassName?: string
+  autoEditing?: boolean
```
ramfox: can we get a comment here about what `autoEditing` means?

src/chrome/DropZone.tsx
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  const { getRootProps, getInputProps, open } = useDropzone({
+    onDrop,
+    noClick: true,
+    noKeyboard: true
+  })
+
+  const handleSelectClick = (e: any) => {
+    console.log('HERE')
```
ramfox: stray log!

src/features/dataset/DatasetHeaderRender.tsx
```diff
@@ -0,0 +1,122 @@
...
+  editable?: boolean
+  // affects the behavior of the title. if true, the user can edit the title inline
+  manualCreation?: boolean
+  // fires when the user renames the dataset
+  onRename?: (_: string, value: string) => void
+  // fires when the user changes the dataset title
+  onTitleChange?: (_: string, value: string) => void
+}
+
+const DatasetHeaderRender: React.FC<DatasetHeaderRenderProps> = ({
```
ramfox: I'm not sure I'm sold on the name, because we've already been using the work `Layout` in other places to do something similar. Does `DatasetHeaderLayout` make sense? If not, should we change our components that use the `Layout` suffix to say `Render` instead (aka `DatasetFixedLayout`, `DatasetScrollLayout`)

src/features/manualDatasetCreation/state/manualDatasetCreationState.ts
```diff
@@ -0,0 +1,71 @@
...
+    state.readme = action.text
+  },
+  SET_MANUAL_DATASET_CREATION_META: (state: ManualDatasetCreationState, action: ManualDatasetCreationSetMetaAction) => {
+    state.meta = action.meta
+  },
+  SET_MANUAL_DATASET_CREATION_TITLE: (state: ManualDatasetCreationState, action: ManualDatasetCreationSetTextAction) => {
+    state.meta.title = action.text
+  },
+  'API_SAVEUPLOAD_REQUEST': (state: ManualDatasetCreationState, action) => {
+    state.fileUploading = true
```
ramfox, line 60: need to clear the `state.uploadError` here

`state.uploadError = ''`

ramfox (changes requested):
A few change requests!

But also, having a hard time reviewing all the `Manual` dataset component files, since there doesn't seem to be a route to view them. That just means that as we expand the "Manual" editing, we just need to be aware that those files need a more indepth look as we refactor.

src/chrome/EditableLabel.tsx
chriswhong: This was intended to put the EditableLabel into edit mode by default, but it's not being used anymore so I removed it.

src/features/dataset/DatasetHeaderRender.tsx
chriswhong: Changed to `DatasetHeaderLayout`

PR #594, opened 2021-11-19
feat: extending e2e tests to clean after themselves

### LINKED ISSUE #319: e2e tests should clean up after themselves, removing datasets and deleting test user
If I understand this correctly, bugs in cloud/backend prevent our current test suite from cleaning up after itself. Let's solve those issues & then update tests so we're ending back in a clean state. e2e tests should still use `test_` username / dataset prefixes for easier identification.

### Steps to Complete
* [ ] Identify issues preventing e2e tests from doing a "clean" setup where no "special" usernames are required, and any issues preventing workflow, dataset, and user deletion. Link those issues here.
* [ ] Fix those issues
* [ ] Update e2e tests
* [ ] party

boandriy:
extending e2e tests to clean after themselves deleting the created datasets

closes #319

### DIFF
--- a/e2e-tests/golden_path.spec.ts
+++ b/e2e-tests/golden_path.spec.ts
@@ -1,16 +1,8 @@
 import { test, expect, Page } from '@playwright/test';
 
 const APP_URL = process.env.TEST_E2E_APP_URL || 'http://localhost:3000';
-const USERNAME = process.env.TEST_E2E_USERNAME || '';
-const USER_PASSWORD = process.env.TEST_E2E_PASSWORD || '';
 
 test.describe.serial('Golden path', () => {
-  test.skip(() => {
-    if (!USERNAME || !USER_PASSWORD)
-      console.log('Please provide credentials!');
-    return !USERNAME || !USER_PASSWORD
-  }, 'Please provide credentials!');
-
   let page: Page;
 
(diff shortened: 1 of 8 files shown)


ramfox:
cool, the dataset clean up looks good!

I just have questions about the username/password & creating a new account.

Is there a plan for how we are going to get to the point of being able to remove a user all in the e2e tests?

Also, it doesn't look like we use the `USERNAME` or `PASSWORD` global vars any longer, we should remove those & the check to ensure that they are set environment variables.

boandriy:
> cool, the dataset clean up looks good!
> 
> I just have questions about the username/password & creating a new account.
> 
> Is there a plan for how we are going to get to the point of being able to remove a user all in the e2e tests?
> 
> Also, it doesn't look like we use the `USERNAME` or `PASSWORD` global vars any longer, we should remove those & the check to ensure that they are set environment variables.

I talked with @Arqu , about cleaning the users. And we decided that this will be done on the BE, as there are few issues with giving an access to just delete a user. Yep, will clean the Unused global vars.

ramfox (approved):
fab! LGTM

Arqu:
Yep, the above is correct. User deletion will be done manually every now and then (will make a cron if it becomes tedious) simply based off of the prefix for the user.

PR #600, opened 2021-11-22
feat: added modal for ds creation

### LINKED ISSUE #464: New Dataset Options
When the user clicks "New Dataset", they should be presented with a couple of options:  Manual Upload or Create from Workflow

We need to present both of these options, and it feels like a modal with some helpful text, graphics, etc can help the user figure out what they want to do.

Even if we aren't ready to add manual upload just yet, we can add the option and disable it.

boandriy:
added modal for ds creation
Added e2e test to support the modal

closes #464


chriswhong (approved):
LGTM, added a commit to re-word the text a bit

ramfox (approved):
Looks great!!

LGTM

PR #603, opened 2021-11-23
feat: add react helmet, add titles, descriptions, etc to all routes

chriswhong:
- adds `react-helmet` and a new `<Head>` component (pulled in from old Qri Cloud) allowing for the addition of `title`, `description`, etc to each route

- pages that should have unfurling links also pass `pathname` to the `Head` component.  We will need to confirm whether this behaves well with our SSR implementation.


src/features/app/Head.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  return (
+    <>
+      <Helmet>
+        <meta name="twitter:card" content="summary" />
+        <meta name="twitter:site" content="@qri_io" />
+        <meta property="og:type" content="website" />
+        <meta property="og:site_name" content="Qri Cloud"/>
+
+        { title && <title>{title}</title> }
```
boandriy, line 31: Can we use 1 && operator and wrap the content with React.Fragment aka <></> ?

boandriy (approved):
LGTM!

src/features/app/Head.tsx
boandriy: Instead of this : 
`
{ title && <title>{title}</title> }
{ title && <meta name="twitter:title" content={title} /> }
{ title && <meta property="og:title" content={title} /> }
`
lets try something like this:
`
{ title && 
<>
  <title>{title}</title>
  <meta name="twitter:title" content={title} />
  <meta property="og:title" content={title} />
</>}
`

src/features/app/Head.tsx
```diff
@@ -0,0 +1,51 @@
...
+        { title && <meta name="twitter:title" content={title} /> }
+        { title && <meta property="og:title" content={title} /> }
```
boandriy:
```suggestion
        { title && <>
          <title>{title}</title>
          <meta name="twitter:title" content={title} />
          <meta property="og:title" content={title} />
        <> }
```

boandriy: Something like this above, I saw a few more examples in the same file.

PR #607, opened 2021-11-24
fix: react-helmet titles not appearing

chriswhong:
Fixes a regressing from #603.  `react-helmet` doesn't like react fragments as children, and was causing the frontend to crash in development.  The production build worked, but the titles and other `HEAD` things that react-helmet was supposed to be assigning were not assigned.  

There will need to be some follow-on work to get the correct `meta` tags to show up when doing server side rendering, but we can tackle that later.


ramfox (approved):
LGTM!

PR #608, opened 2021-11-25
feat: run queue

### LINKED ISSUE #537: [workflow]: User can see that a run has been requested but hasn't started (run queue)
When the user clicks run now, or dry run, we need a queued status and associated UI.  This could be a slower subdued spinner, with a tooltip that says "waiting..." or something similar, then when the run starts it can be brighter/faster/more prominent.

boandriy:
added run queue, both on the collection and run now page. 
closes #537

### DIFF
--- a/src/features/activityFeed/ActivityList.tsx
+++ b/src/features/activityFeed/ActivityList.tsx
@@ -7,12 +7,14 @@ import RelativeTimestamp from '../../chrome/RelativeTimestamp'
 import Icon from '../../chrome/Icon'
 import DatasetCommitInfo from '../../chrome/DatasetCommitInfo'
 import RunStatusBadge from '../run/RunStatusBadge'
-import { LogItem } from '../../qri/log'
+import { VersionInfo } from '../../qri/log'
 import { customStyles, customSortIcon } from '../../features/collection/CollectionTable'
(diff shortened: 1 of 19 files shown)


chriswhong (approved):
LGTM with one nit:

"Waiting" should be capitalized so is looks the same as "Running" and "Success" and "No Changes" when showing run statuses

ramfox:
Hey, so examining your PR and taking a look at the backend.

I think the way you are adding the "Waiting" event is a clever way to solve this. But it means that we lose the ability to "cancel" the queued run, because you need the `RunID` in order to cancel the event. So I took a look at what we were sending over the websocket:

```
{ 
    // this is the run id
    data: "6ea4d6e2-e145-44e3-b8f2-a8c7988140a9"
    // this is ALSO the run id
    sessionID: "6ea4d6e2-e145-44e3-b8f2-a8c7988140a9"
    ts: 1639068512035200300
    type: "automation:RunQueuePush"
}
```

This is NOT ENOUGH INFO!!! How are you supposed to relate that run id to the dataset? I keep failing to think of the practical way this is going to be used in the frontend.

So, I'm going to refactor the backend to send over the dataset id (InitID) as well. Let's have a quick meeting tomorrow where we talk about the best way this can work for you in frontend.

PR #609, opened 2021-11-27
refactor(UserProfile): request a user's datasets from the core `/list` endpoint

ramfox:
Now that we have `Collection` in core and cloud & now that we have basic sorting on the `/list` endpoint, we can remove references of the `/dataset_summary` cloud specific endpoint.

NOTE

This will only work on cloud when PR https://github.com/qri-io/qri/pull/1981 is merged & then cloud dependencies are updated.

We may want to hold off on merging this PR until those conditions are met, otherwise sorting on the user profile page will be broken on cloud.

### DIFF
--- a/src/features/userProfile/state/userProfileActions.ts
+++ b/src/features/userProfile/state/userProfileActions.ts
@@ -1,13 +1,34 @@
+import { ThunkDispatch } from 'redux-thunk'
+
+import { RootState } from "../../../store/store"
 import { ApiAction, ApiActionThunk, CALL_API } from "../../../store/api"
 
 import { NewSearchResult } from '../../../qri/search'
-import { UserProfileDatasetListParams } from '../../../qri/userProfile'
+import { UserProfile, UserProfileDatasetListParams } from '../../../qri/userProfile'
+import { mapVersionInfo } from "../../collection/state/collectionActions"
+import { USERPROFILE_SET } from "./userProfileState"
 
 export const searchPageSizeDefault = 25
 
-export function loadUserProfile (username: string): ApiActionThunk {
-  return async (dispatch, getState) => {
+export function loadUserProfile (username: string) {
+  return async (dispatch: ThunkDispatch<any, any, any>, getState: () => RootState) => {
+    const sessionUser = getState().session.user
+    if (sessionUser.username === username) {
+      return dispatch(setUserProfile(sessionUser))
+    }
     return dispatch(fetchUserProfile(username))
+  }
+}
+
+export interface UserProfileAction {
+  type: string
+  user: UserProfile
+}
+
+function setUserProfile (user: UserProfile): UserProfileAction {
+  return {
+    type: USERPROFILE_SET,
+    user
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/qri/userProfile.ts
```diff
@@ -15,37 +15,53 @@ export interface UserProfile {
...
+  networkAddrs: string[]
   id: string
   currentKey: string
-  EmailConfirmed: boolean
+  emailConfirmed: boolean
   isAdmin: boolean
 }
 
-export const NewUserProfile = () => {
+export const NewUserProfile = (d: Record<string, any>): UserProfile => {
```
ramfox, line 26: We have a similar `NewDataset` function that I've found useful. Converted this to work in the same manner.

src/qri/userProfile.ts
```diff
@@ -15,37 +15,53 @@ export interface UserProfile {
...
+  } else if (d.created) {
+    created = d.created
+  }
+
+  if (typeof d.updated === 'string') {
+    const u = new Date(updated)
+    updated = u.getTime() / 1000
+  } else if (d.updated) {
+    updated = d.updated
+  }
```
ramfox, line 42: the `UserProfile` coming back from cloud has the updated and created fields as numbers, backend has them coming back as strings. This unifies!

src/features/userProfile/state/userProfileState.ts
```diff
@@ -62,20 +65,25 @@ export const userProfileReducer = createReducer(initialState, {
...
+  },
+
   'API_USERPROFILEDATASETS_REQUEST': (state: UserProfileState, action) => {
     state.datasets.loading = true
   },
   'API_USERPROFILEDATASETS_SUCCESS': (state: UserProfileState, action) => {
     state.datasets.results = action.payload.data
-    state.datasets.pageInfo = action.payload.pagination
+    state.datasets.pageInfo = action.payload.request.pageInfo
+    state.datasets.pageInfo.resultCount = state.datasets.results.length
```
ramfox, line 86: We are now using the `list` endpoint, that does not respond with the same kind of pageInfo. We are instead creating it here to match the expectations of the `UserProfile` page. Also, working on unifying our pagination, so this will probably change soon.

src/features/userProfile/state/userProfileActions.ts
```diff
@@ -1,16 +1,37 @@
...
 export const searchPageSizeDefault = 25
 
-export function loadUserProfile (username: string): ApiActionThunk {
-  return async (dispatch, getState) => {
+export function loadUserProfile (username: string) {
+  return async (dispatch: ThunkDispatch<any, any, any>, getState: () => RootState) => {
+    const sessionUser = getState().session.user
+    if (sessionUser.username === username) {
+      return dispatch(setUserProfile(sessionUser))
+    }
```
ramfox, line 18: This is both to prevent us from double fetching, but also a work-around if you are working locally. If you are working locally, you may not have a cloud account, which means calling `fetchUserProfile` will come back with a 404 error.

chriswhong (approved):
Tried this locally and the user profile loaded without any issues.  I was able to use the sorting UI and got the expected results. LGTM!

PR #610, opened 2021-11-27
fix(NoBody): show `NoBody` component if a dataset has no body

### LINKED ISSUE #593: [History and Preview] Empty body
When you are uploading a dataset and you deploy without running the body could be empty and the default state on the preview is loading dots.  This also happens on the history page.

What you should see is this

Screenshot 2021-11-18 at 22 03 02

ramfox:
closes #593 

Also removes extraneous call to get the dataset body,  now that (as per https://github.com/qri-io/qri/pull/1980, when it is merged) the `/ds/get` endpoint returns a dataset preview if no selectors are given.

NOTE

Until the above pr is merged and then cloud is updated to use that version of qri, this PR will cause problems on the dataset preview page. We may want to wait for the backend to be pushed before merging this pr.

### DIFF
--- a/src/qri/dataset.ts
+++ b/src/qri/dataset.ts
@@ -92,6 +92,27 @@ export type ComponentName =
  | 'viz'
  | 'stats'
 
+// rankedComponentNames is a ranked list ranging from the components we deem
+// it most important that user see first, to the least important
+const rankedComponentNames = ['body', 'readme', 'transform', 'viz', 'meta', 'stats', 'commit', 'structure']
+
+export function getRankedComponentNames (dataset: Dataset): ComponentName[] {
+  return rankedComponentNames.reduce<ComponentName[]>((acc, name) => {
+    const comp = getComponentFromDatasetByName(dataset, name)
+    if (name === "body" && comp) {
+      const bodyComp = comp as BodyComponent
+      if (bodyComp.body) {
+        acc.push(name as ComponentName)
+      }
+      return acc
+    }
+    if (comp) {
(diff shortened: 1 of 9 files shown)


chriswhong (changes requested):
For the tabbed view: 
I don't think we should have an active tab with nothing in it.  If there is no body, the body tab should be inactive and the first tab for a component that exists should be selected by default.

For the preview:
I think a message is appropriate to show when previewing a dataset with no body, but the wording should be third person and we should avoid using the term `body` as we have replaced it elsewhere in the UI.   "No data"  (New line, in gray:)"The dataset owner hasn't added any data".

src/qri/dataset.ts
```diff
@@ -92,6 +92,27 @@ export type ComponentName =
  | 'viz'
  | 'stats'
 
+// rankedComponentNames is a ranked list ranging from the components we deem
+// it most important that user see first, to the least important
+const rankedComponentNames = ['body', 'readme', 'transform', 'viz', 'meta', 'stats', 'commit', 'structure']
```
ramfox, line 97: @chriswhong what do you think of this ranking? Please let me know what you think the order should be if your thoughts differ!

chriswhong: I don't have strong opinions on this.  It's fine for now, we can tweak it later if necessary.

chriswhong (approved):
🚀 LGTM.  Pushed a commit to modify the NoBody text, everything else looks grand.

PR #614, opened 2021-11-29
fix: adjust user avatar location

### LINKED ISSUE #606: [Profile] Avatar position
The position of the avatar/profile image it shouldn't be out of the white box as you can see here:

Screenshot 2021-11-24 at 10 26 29

It should be like this:

Screenshot 2021-11-24 at 11 08 10

chriswhong:
image

Style tweaks to make the user avatar appear in the correct place.

Closes #606


tefyzambrano (commented):
Great!

PR #615, opened 2021-11-29
feat: redirect to / on log out

### LINKED ISSUE #605: [Profile] Log out state
When I'm on the profile page and decide to log out, it stays on the same page. I think you should be send to the product landing page

Screenshot 2021-11-24 at 10 27 05
e.

chriswhong:
Adds a redirect to `/` when user manually logs out from the user session menu.   Closes #605


boandriy (approved):
LGTM!

PR #626, opened 2021-12-01
feat: update SSR to include twitter card and og tags

### LINKED ISSUE #584: [SSR] Add Link Unfurling
We added react helmet in #603, but the `og` and other meta tags don't come through correctly in SSR. 

In SSR, we are already injecting jsonLD  into `index.html` here: https://github.com/qri-io/frontend/blob/7bf4732f7d75d85c6338054f2079fe821e8e3794/server/index.js#L97-L104

We can use the same call to the preview to populate `og` tags and whatever else is necessary for link unfurling.

chriswhong:
- Adds some generic `meta` tags to `index.html` to support link unfurling/twitter cards
- Renames the SSR util `composeHeadData` to `composeHeadTags`, it now returns `<title>` and various `<meta>` tags needed for link unfurling/twitter cards, with the title and dataset for the selected dataset injected to the correct place.
Slack___twitter___qri___100_new_items

Closes #584


boandriy (approved):
LGTM!

PR #627, opened 2021-12-02
fix: running event is now removed from the run log

### LINKED ISSUE #621: Run Now in Run Log shows "Running" forever when the run has already completed
chriswhong_popular-baby-names_run_log___Qri

I am seeing "Running..." as the run status after run now, even though the toast tells me the run has completed.  If I reload, it shows as completed.  I think we are missing a websocket update or something.

boandriy:
running event is now removed from the run log

fixes #621


ramfox (approved):
LGTM

PR #631, opened 2021-12-06
feat: dataset editing + new dataset convergence

### LINKED ISSUE #273: Manual Edit UI for Readme and Meta
The manual dataset uploading UI is the same layout as the dataset editor.  Since that work is already in, let's modify this work to use those components.

**Add meta and readme editing to manual dataset creation flow**
- [ ] Make readme and meta tabs in the manual update screen editable

**Create a new `/edit` route with the same layout as the manual upload route**
- [ ] When the user clicks 'edit' in the history view, let's navigate to `/:username/:datasetname/edit` and show the same UI that we show for manual dataset creation, but with different state for editing the active dataset

- [ ] Confirm that the full editing flow works
  - [ ] Go to existing dataset history view
  - [ ] Click "edit"
  - [ ] Navigate to `/:username/:datasetname/edit`
  - [ ] Make a change to the readme and/or metadata
  - [ ] Make a change to the Title in the dataset header, it should also be reflected in the meta tab
  - [ ] Commit
  - [ ] Navigate back to `/history` showing the current version of the dataset

boandriy:
added functionality to edit Meta and Readme of a dataset

closes #273


chriswhong:
Added a commit with a major refactor, consolidating the previous work done on this branch to edit readme and meta for an existing dataset with the new dataset creation flow that was recently merged into master.

Both new dataset creation and existing dataset are now consolidated into the same feature in `/src/datasetEditor`.  There are two top-level components, `NewDatasetEditor` and `ExistingDatasetEditor`, both of which render `DatasetEditorLayout` in different ways for their respective UXs.  

Both also share the same state, which includes a `Dataset` which will be mutated when the user edits components, loading and error booleans, and a `File` for body updates

```
export interface DatasetEditorState {
  dataset: Dataset
  loading: boolean
  error: string
  // file is used to store the file reference when POSTing a body via form data
  file?: File
}
```

There is a lot more to do, but as of dc08ea8 the user can create a new dataset from a CSV, and modify the readme and meta before committing.  They can also edit existing datasets via a button in the history list.

Still TODO:

*PUNT*
- [ ] We should only show the commit bar when there are changes.  This involves comparing the workingDataset with the original.  We can punt on this if it's too difficult, but it would be a nice effect to have the commit bar pop up as soon as the first edit happens.  (status quo: in `NewDatasetEditor`, it appears when the file is added, which might be what we want as it forces the user to have a body.  In `ExistingDatasetEditor` the prop is hard coded to `true` to it always shows.

- [x] Dataset Title Editing Behavior - In `NewDatasetEditor`, the dataset name is stored in component state and used when making the api call to create the new dataset.  However, in `ExistingDatasetEditor` we aren't currently handling a change to the dataset name.  Changing the name via the same method we had enabled in the past would require a trip to the server, and I think this would make the user lose their changes unless we add some code to prevent this.  Should the user be able to change the name of a dataset they are editing?  I tend to think we should disable the name edit field, and make a name change be a bigger deal with its own modal or something.  We need the name editor for new datasets for obvious reasons, but I think it's less necessary when editing an existing dataset.

- [x] Error handling for the commit api call in `ExistingDatasetEditor`.  We show a loading spinner in the commit button, but don't handle the error or display it.

- [x] User should be able to update the body of an existing dataset by dragging in a new file.  I think this won't work with the `commit` api call we are doing at the end of the `ExistingDatasetEditor` flow, so we should check with @Arqu to see if it's possible ATM.

*PUNT*
- [ ] We have existing components for `MetaEditor` and `ReadmeEditor`, but we build the editors into `Meta` and `Readme`.  I like the pattern of keeping these separate and just calling a different component for editing.  If it's not too much work lets separate them again.

Follow-on issues
- Add a structure editor

package.json
```diff
@@ -123,6 +123,7 @@
       "@typescript-eslint/no-empty-function": "off",
       "@typescript-eslint/prefer-readonly": "off",
       "@typescript-eslint/no-var-requires": "off",
+      "@typescript-eslint/no-unused-vars": "warn",
```
ramfox: Personally think we should be strict about `no-unused-vars`, is there a particular case we are coming up against?

src/chrome/DatasetCommitInfo.tsx
```diff
@@ -38,7 +38,6 @@ const DatasetCommitInfo: React.FC<DatasetCommitInfoProps> = ({
         'font-semibold': !inRow
       })}>
         <div className={classNames(`dataset_commit_info_text truncate transition-colors flex-grow ${hover && 'group-hover:underline group-hover:text-qripink-600'}`, {
-
```
ramfox, line 41: dunno about this one, looks like a really valuable line of code you deleted here

src/features/datasetEditor/ExistingDatasetEditor.tsx
```diff
@@ -0,0 +1,89 @@
...
+    })
+
+    const theDataset = NewDataset({
+      meta: dataset.meta,
+      readme: dataset.readme
+    })
+
+    commitDataset(qriRef, theDataset, commitTitle, file)(dispatch)
+      .then(res => {
+        console.log('here', res)
```
ramfox: stray log

src/features/datasetEditor/DatasetEditorLayout.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const handleCloseClick = () => {
+    if (onClose) { onClose() }
+  }
+
+  return (
+    /* top level div is a clone of DatasetFixedLayout, but we didn't use it because we need a different Header */
+    <div className='flex flex-col flex-grow bg-qrigray-100' style={{ borderTopLeftRadius: 20 }}>
+      <div className='flex flex-col overflow-hidden p-7 flex-grow border-qritile-600 border-2' style={{ borderTopLeftRadius: 20 }}>
+        <Head data={{
+          title: `${dataset.meta?.title} - New Qri Dataset`
```
ramfox: if there is no `meta.title` this reads `undefined - New Qri Dataset` (showing up on an older-ish dataset I have)

maybe we want `${dataset.meta?.title || dataset.name} - New Qri Dataset`, just to be on the safe side?

Also, this language `New Qri Dataset` shows up for both the existing & new versions of the editor. Not sure if this is on purpose.

src/features/datasetEditor/state/datasetEditorActions.ts
```diff
@@ -0,0 +1,113 @@
...
+//         method: 'POST',
+//         body: {
+//           ref: `${qriRef.username}/${qriRef.name}`,
+//           dataset: dataset
+//         }
+//       }
+//     }
+//     return dispatch(action)
+//   }
+// }
```
ramfox: either remove or add a comment about when we should be able to comment this back in

src/features/dsComponents/meta/MetaEditor.tsx
```diff
@@ -234,9 +234,8 @@ const MetaEditor: React.FC<MetaEditorProps> = ({
               value={data[field.field]}
               placeholder={field.placeholder}
               maxLength={field.maxLength}
-              onChange={(e: React.FormEvent) => {
-                const target = e.target as HTMLInputElement
-                handleWrite(field.field, target.value)
+              onChange={(d: string) => {
+                handleWrite(field.field, d)
```
ramfox, line 238: this is a more general comment about the `DatasetEditor` & its children components (`TabbedDatasetEditor`, `ComponentEditor`, `MetaEditor` & `ReadmeEditor`), they don't seem to be used any longer. If they aren't being kept around for future use, we should remove them in this pr or a follow up!

src/features/commits/DatasetCommitList.tsx
```diff
@@ -45,6 +47,19 @@ const DatasetCommits: React.FC<DatasetCommitsProps> = ({
...
+        <li className='flex items-stretch text-black tracking-wider'>
+          <div className='relative w-4 mr-5 flex-shrink-0'>
+            <div className={classNames('absolute top-2.5 w-4 h-4 rounded-3xl border border-qritile-600')}>&nbsp;</div>
+            <div className='relative line-container w-0.5 mx-auto h-full'>
+              <div className='absolute -bottom-2 w-full h-11 bg-gray-300 rounded'>&nbsp;</div>
+            </div>
+          </div>
+          <div className='w-full'>
+            <Link to={`/${qriRef.username}/${qriRef.name}/edit`}>
+              <Button className='mb-6' block disabled={loading || (commits.length ? path !== commits[0].path : false)} type='primary-outline' icon='edit'>Edit</Button>
```
ramfox: we should have some check here if the user is logged in and it is their dataset, only then do we should the `Edit` button

src/features/commits/DatasetCommitList.tsx
```diff
@@ -56,17 +71,19 @@ const DatasetCommits: React.FC<DatasetCommitsProps> = ({
...
-            />
-          ))}
+          : <>
+            {commits.map((logItem, i) => (
+              <DatasetCommitListItem
+                key={i}
+                loading={false}
+                logItem={logItem}
+                active={logItem.path === path}
+                // first={i === 0 && !editable} (restore when there is <NewVersionButton> at the top of the list)
```
ramfox: If I'm translating correctly, `NewVersionButton` would be the `Edit` button we now have. I think we can remove this comment as it is no longer relevant.

src/features/dataset/CompletionCheckmark.tsx
```diff
@@ -0,0 +1,29 @@
+import React from "react"
+
+interface CompletionChenckmarkProps {
+  type: 'green' | 'gray' | 'orange' | 'outline'
+  className?: string
+}
+
+const CompletionCheckmark: React.FC<CompletionChenckmarkProps> = ({ type, className }) => {
```
ramfox, line 8: 2 things: 
1) Is this used anywhere?
2) we seem to have two components called `CompletionCheckmark`, one defined in `src/features/dataset` & one defined in `src/features/datasetEditor`

src/features/dataset/DatasetHeaderLayout.tsx
```diff
@@ -28,25 +28,28 @@ export interface DatasetHeaderLayoutProps {
...
+  nameEditable?: boolean
   // fires when the user renames the dataset
-  onRename?: (_: string, value: string) => void
+  onNameChange?: (_: string, value: string) => void
+  // allows editing the title, which will call onTitleChange
+  titleEditable?: boolean
   // fires when the user changes the dataset title
   onTitleChange?: (_: string, value: string) => void
+  // whether the user owns the dataset
+  userOwnsDataset?: boolean
```
ramfox: I like this naming! Wouldn't be opposed to changing `editable` to `userOwnsDataset` in other places that are appropriate (not necessary for this PR, just making a note as we move forward)

src/features/dataset/DatasetNameCommitModal.tsx
```diff
@@ -16,12 +16,12 @@ import { ModalType, selectNavExpanded } from "../app/state/appState"
 import { QriRef, newQriRef } from "../../qri/ref"
 import { loadDatasetCommits } from "../commits/state/commitActions"
 
-export interface DatasetHeaderLayoutProps {
+export interface DatasetNameCommitModalProps {
```
ramfox: Is this modal for a future flow?

src/features/dataset/modal/EditDatasetTitleModal.tsx
```diff
@@ -15,8 +15,8 @@ export interface EditDatasetTitleModalProps {
 
 const EditDatasetTitleModal: React.FC<EditDatasetTitleModalProps> = ({ title, onCommit }) => {
   const dispatch = useDispatch()
-  const titleEditError = useSelector(selectTitleError)
-  const loading = useSelector(selectIsDatasetLoading)
+  const titleEditError = ''
+  const loading = ''
```
ramfox: I don't understand this change, either it should change back or could use a comment clarifying the intention!

src/features/dataset/state/datasetActions.ts
```diff
@@ -190,6 +194,54 @@ export interface ResetDatasetStateAction {
...
+  }
+}
+
+export function setDatasetReadme (readme: Readme): SetDatasetReadmeAction {
+  return {
+    type: SET_DATASET_README,
+    readme
+  }
+}
+
```
ramfox: It looks like most of these may have been replaced by the action in `src/features/datasetEditor/state/datasetEditorActions`

Unless they are being saved for future purposes, we should remove them.

src/features/dataset/state/datasetState.ts
```diff
@@ -18,6 +21,14 @@ export const SET_BODY_LOADING = 'SET_BODY_LOADING'
...
 export const RESET_DATASET_TITLE_ERROR = 'RESET_DATASET_TITLE_ERROR'
 
+export const SET_DATASET_IS_EDITABLE = 'SET_DATASET_IS_EDITABLE'
+
+export const SET_DATASET_META = 'SET_DATASET_META'
+
+export const SET_DATASET_README = 'SET_DATASET_README'
+
+export const SET_DATASET_COMMIT_TITLE = 'SET_DATASET_COMMIT_TITLE'
+
```
ramfox: I think once we get rid of the actions in the above file, we can probably get rid of these as well.

ramfox (changes requested):
This is looking great! Very fun to use!!

I noticed 1 backend specific issue while playing around, documented here: https://github.com/qri-io/qri/issues/1997. If you try to save a new version of a dataset, and only change an existing list of keywords, it will erroneous return "saving failed: no changes", even though the frontend is sending over all of the correct information.

The only two "bugs" I've found in the UI:
1) All users can see the "edit" button for all datasets, even ones they do not own. I think this should be fixed before we merge.
2) The documented bug that has been punted surrounding the commit bar, which sounds totally fine to punt, since pushing the button while having no changes does not cause any issues

I've also requested some changes, mostly around removing code that isn't being used.

Otherwise, looks fab!!

src/features/datasetEditor/DatasetEditorLayout.tsx
chriswhong: I moved the `<Head>` tags up to the respective wrapping components, and added the appropriate title for each along with the OR you suggested.

src/features/dataset/DatasetNameCommitModal.tsx
chriswhong: This is actually that modal that we had in place to edit the title with a commit button, but we've decided to not use it for now.  The component is still around if we choose to use it again.

src/features/dataset/CompletionCheckmark.tsx
chriswhong: I believe these are for the "tic tacs" for showing which components have changed.  We aren't yet ready to use these but we can clean this up later

src/features/dataset/modal/EditDatasetTitleModal.tsx
chriswhong: These were changed as the actions they use are no longer present and it was causing compiling errors.  I just removed the component as we are no longer using it.

src/features/dsComponents/meta/MetaEditor.tsx
chriswhong: There may be bits in here we want to re-use, I'll make an issue for cleanup of the old editor components

PR #636, opened 2021-12-07
Cw/default template

chriswhong:
- Simplifies the default template code in automations to include only a CSV download and commit.
- Refactors template logic so that the workflow editor will always show a template if there is no automation for the current dataset


boandriy (approved):
LGTM!

PR #641, opened 2021-12-10
feat: add loading spinner to run log

### LINKED ISSUE #633: [runlog] add loading state for run log
The run log shows "There are no records to display" when loading.  

When we are loading the collection list, we show the animated bouncing balls as a spinner.  Let's do the same with the activity list to be consistent.

chriswhong:
Adds the same loading spinner to run log that we use on the collection list.

Closes #633


b5 (approved):
LGTM

PR #650, opened 2021-12-17
Cw/workflow refactor

### LINKED ISSUE #616: [Automation] Workflow UX Refactor
Qrimatic_2_0_NEW_–_Figma

- [x] Rename Workflow Nav Item to “Automation”
- [x] Reword Automation Splash Carousel Text: "Get Started with Qri Automated Datasets"
- [x] Reword supporting text on each carousel pane as needed
- [x] Move triggers to sidebar
- [x] Remove deploy button from header, replace with commit bar which pops up from the bottom when user is ready to commit (same as used in dataset editing)
- [x] If editing existing, Commit bar should include "Run Now" checkbox, checked by default
- [x] If new automation, show modal on commit
- [x] On success, the large button in the modal should go to the new dataset's preview page ("See Dataset Preview")

*punt*
- [ ] Modal should clearly show the progression Saving Changes -> Queueing -> Running -> Success (with confetti)

chriswhong:
Refactors the workflow creation and editing UI, Closes #616 

Automation happens via two components `NewAutomationEditor` and `ExistingAutomationEditor` which are children of their respective routes.  Both of these render `WorkflowPage`, which incorporates the same `EditorLayout` we are using in the static dataset editor UI.

`WorkflowPage` then renders `WorkflowEditor`, which contains the code blocks, new sidebar, etc.  

We have replaced "workflow" in the UI, but it is still present in the codebase.  Should be seek to change all references to "workflow" into "automation"?  This would carry over into the backend as well, and may not be worth the effort since workflow is still a thing in the behind the scenes.

In addition to reviewing the code, please pull this down and use it.  Try to go through both creation of an automated dataset and editing the automation of an existing dataset.

TODO:
- [x] Require change to dataset name before proceeding when creating a new automated dataset
- [x] Always show input outlines in editors

Found during review:
- [x] If opening the automation editor on a dataset that doesn't have one, show the template code
- [x] If an existing workflow has multiple steps, combine them all into a single step so only one cell is shown
- [x] When the sidebar is collapsed the (+) add dataset button does not open the new dataset modal
- [x] Handle user removing dataset name completely - on blur, change it back to whatever it was (Edit: we got around this by requiring a name change before enabling the commit button)


package.json
```diff
@@ -123,6 +123,7 @@
       "@typescript-eslint/no-empty-function": "off",
       "@typescript-eslint/prefer-readonly": "off",
       "@typescript-eslint/no-var-requires": "off",
+      "@typescript-eslint/no-unused-vars": "warn",
```
ramfox: this snuck back in!

src/features/workflow/modal/DeployModal.tsx
```diff
@@ -1,35 +1,38 @@
...
-const DEBOUNCE_TIMER = 500
+export interface DeployModalProps {
+  username: string
+  name: string
+  runNow: boolean
+  isNew: boolean
+}
 
-const DeployModal: React.FC = () => {
+const DeployModal: React.FC<DeployModalProps> = ({
```
ramfox: I could go for a name change here, since this is being used after commits and deploy's. `CommitModal`? `ConfirmCommitModal`?

src/features/dataset/modal/EditDatasetTitleModal.tsx
```diff
@@ -0,0 +1,62 @@
...
+import TextInput from "../../../chrome/forms/TextInput"
+// import { selectIsDatasetLoading, selectTitleError } from "../state/datasetState"
+import Spinner from "../../../chrome/Spinner"
+
+export interface EditDatasetTitleModalProps {
+  title: string
+  onCommit: (title: string, commitTitle: string) => void
+}
+
+const EditDatasetTitleModal: React.FC<EditDatasetTitleModalProps> = ({ title, onCommit }) => {
```
ramfox: Is there a plan to use this in the future?

src/features/dataset/DatasetNameCommitModal.tsx
```diff
@@ -0,0 +1,72 @@
+// this component contains a the modal for editing a dataset's title and creating a commit for the change
+// we are going to deprecate it in favor of the meta editor, but we may want to bring it back someday
```
ramfox: if this is planning to be used for the `meta.title` & not the dataset name, we should rename this component

src/chrome/CommitBar.tsx
```diff
@@ -0,0 +1,65 @@
...
+  onCommit: () => void
+  // whether to show the Run now checkbox
+  showRunNow?: boolean
+  // the value of the Run now checkbox
+  runNow?: boolean
+  // fired when the Run now checkbox is changed
+  onRunNowChange?: () => void
+}
+
+const CommitBar: React.FC<CommitBarProps> = ({
```
ramfox, line 29: We should reuse this component in `DatasetEditorLayout`

ramfox (changes requested):
looking good!

We talk about some bugs/changes in chat already, but here are some (very few) notes!

src/features/workflow/modal/DeployModal.tsx
chriswhong: Renamed to AutomationCommitModal

src/features/dataset/DatasetNameCommitModal.tsx
chriswhong: I removed this, we aren't using it ATM.

src/chrome/CommitBar.tsx
chriswhong: Great catch!

src/features/dataset/modal/EditDatasetTitleModal.tsx
chriswhong: Doubtful. I removed it

chriswhong:
- `DeployModal` and `SplashModal` are renamed `AutomationCommitModal` and `AutomationSplashModal`
- The Automation editor will show the splash modal on an existing dataset that does not yet have a transform.
- Removed the the reducer `API_PREVIEW_SUCCESS` from workflowState, this was used to set `workflow.dataset` to be the active dataset but was causing problems when a dataset didn't have a transform.  Instead `ExistingAutomationEditor` now checks for the transform, and if one doesn't exist, it dispatches the new `setWorkflowDataset` action and adds the template code.
- Existing transforms that had multiple steps will have all of the steps reduced into one in the UI (and will thus have only one step on the next commit), making the new UI compatible with earlier step-based transforms.
- `CommitBar` component is now used in all editors
- On New Automations, the name and title inputs will always show an outline, making them more form-like versus editor-like.
- Fixed new dataset button bug (no action on click when collapsed)
- Removes unused automation template code

ramfox (approved):
Wooo! This looks great!!

One small bug (I think), but then it looks good to merge.

This might be on purpose, but I can only change the dataset name when the dataset is new. On the "edit" and "automation" pages, after the dataset has already been created, I can't edit the name.

PR #659, opened 2021-12-22
refactor: eliminate unnecessary calls to /ds/get

### LINKED ISSUE #266: Rendering a Preview requires multiple API calls
`/ds/get` should include a body with 100 rows.  Right now we have to make multiple calls to render the dataset preview.

chriswhong:
Closes #266 

We had api calls spread across `DatasetRoutes`, `DatasetWrapper`, and `DatasetPreviewPage`, resulting in 3 separate calls to the same endpoint (`/ds/get`) when a preview route was loaded.

- We have separate state for `dataset` and `preview` but they are both populated by the same data, so this refactors the reducers for both to listen for the same action.
-  Calls happening in `DatasetRoutes` are moved down into `DatasetWrapper` so they are all happening in the same place, leaving `DatasetRoutes` to just be a router, not a router and a data-getter.

### DIFF
--- a/src/features/dataset/DatasetRoutes.tsx
+++ b/src/features/dataset/DatasetRoutes.tsx
@@ -1,14 +1,6 @@
-// All dataset routes share DatasetWrapper, which handles the rendering of the
-// dataset menu and fetches dsPreview (the latest version of the dataset)
-// All routes use dsPreview to render the dataset header, so it is always needed
-// regardless of th other dataset content being displayed
-
-// DatasetPreviewPage fetches the other necessary parts of the preview (body + readme)
-
-import React, { useEffect } from 'react'
+import React from 'react'
 import { Route, Switch, useParams } from 'react-router'
 import { useLocation } from 'react-router-dom'
-import { useDispatch } from "react-redux"
 
 import ExistingAutomationEditor from '../workflow/ExistingAutomationEditor'
 import DatasetComponents from '../dsComponents/DatasetComponents'
@@ -17,7 +9,6 @@ import DatasetPreviewPage from '../dsPreview/DatasetPreviewPage'
 import DatasetIssues from '../issues/DatasetIssues'
 import { newQriRef } from '../../qri/ref'
 import DatasetWrapper from '../dsComponents/DatasetWrapper'
-import { loadDataset, loadHeader } from "./state/datasetActions"
 import { PrivateRoute } from '../../routes'
 import ExistingDatasetEditor from "../../features/datasetEditor/ExistingDatasetEditor"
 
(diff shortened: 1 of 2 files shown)


src/features/dsPreview/state/dsPreviewState.ts
```diff
@@ -29,16 +29,16 @@ const initialState: DsPreviewState = {
 }
 
 export const dsPreviewReducer = createReducer(initialState, {
-  'API_PREVIEW_REQUEST': (state) => {
+  'API_DATASET_REQUEST': (state) => {
```
ramfox: Ahh, so this is intro-ing a bug.

We actually want the `preview` to be a separate call: the preview should always be the latest version.

Since we call `loadDataset` when we iterate over each history, that means this `preview` here will be over-ridden by whatever the last version of the dataset we were looking at in the "history" page.

src/features/dsPreview/DatasetPreviewPage.tsx
```diff
@@ -57,11 +55,6 @@ const DatasetPreviewPage: React.FC<DatasetPreviewPageProps> = ({
     readmeContainerHeight = 'auto'
   }
 
-  useEffect(() => {
-    const ref = newQriRef({ username: qriRef.username, name: qriRef.name, path: qriRef.path })
-    dispatch(loadDsPreview(ref))
```
ramfox: mentioned in other comments, but we need to keep `loadDsPreview`. We can maybe come up with different metrics for when we go to fetch, though.

What are circumstances in which the "head" dataset would change?

1) if we are loading the page for the first time and no preview exists
2) if we are loading the page for that dataset for the first time (so different username/name combo. This may also get triggered if the dataset gets renamed, but I think that's fine)
3) if we have added a new version to the dataset. This can happen via the automation side when we commit a new version of the transform, if an automation runs, or if we do a manual update. In all three cases, we can potentially leverage the websocket events to know that a save occurred. If the save occurred, and we are currently viewing that dataset, we can trigger a `loadDsPreview`, to get the latest "head" dataset.

OR

we just go back to loading it every time we visit the page.

ramfox (changes requested):
Into pushing the ds loading down into the wrapper, rather than the router, that seems v sane.

But removing some of the `loadDsPreview` stuff is intro-ing a bug.

chriswhong:
I pushed a commit which:
- restores dispatching `fetchDsPreview` in `DatasetPreviewPage`
- restores `API_PREVIEW_XXX` reducers in `dsPreviewState`
- removes `loadDataset` from `DatasetWrapper`

This way, we have everything only fetching data when it needs to:

- `DatasetWrapper` gets the header, workflow, and runlog.  These are what is needed for all dataset routes. (ideally we would be able to determine in the call to header whether there is a workflow and run activity so these calls can be moved down to their respective route components, but we can do that later)
- `DatasetPreviewPage` fetches the preview
- `DatasetComponents` fetches whatever dataset version is specified.

I think #3 you mentioned above is a separate issue, we should make sure preview is updated after a commit.

ramfox:
> * `DatasetWrapper` gets the header, workflow, and runlog.  These are what is needed for all dataset routes. (ideally we would be able to determine in the call to header whether there is a workflow and run activity so these calls can be moved down to their respective route components, but we can do that later)

We actually should be able to determine from `header` if a dataset has a workflow, if the `VersionInfo` has a non-empty `WorkflowID` field, than there is a workflow & run activity!

> I think #3 you mentioned above is a separate issue, we should make sure preview is updated after a commit.

Because of the way the `Preview` page is currently loaded, it will fetch the preview each time you switch back to that component, so we have our butts covered if there is a commit.

ramfox (approved):
> * `DatasetWrapper` gets the header, workflow, and runlog.  These are what is needed for all dataset routes. (ideally we would be able to determine in the call to header whether there is a workflow and run activity so these calls can be moved down to their respective route components, but we can do that later)

We actually should be able to determine from `header` if a dataset has a workflow, if the `VersionInfo` has a non-empty `WorkflowID` field, than there is a workflow & run activity!

> I think #3 you mentioned above is a separate issue, we should make sure preview is updated after a commit.

Because of the way the `Preview` page is currently loaded, it will fetch the preview each time you switch back to that component, so we have our butts covered if there is a commit.


But this LGTM!!!

PR #660, opened 2021-12-24
feat: add virtualized body scrolling

### LINKED ISSUE #158: [body] Add virtualized body scrolling
When viewing full body in the history view, users should be able to scroll through the entire body.  At the moment we are only showing one page and have no way for the user to see more data.

chriswhong:
image

Adds `react-virtualized` to `BodyTable` (Closes #158)

- Previews use the old `BodyTable` component which is renamed to `BodyPreviewTable`.  We probably want to consolidate these, but I left the old one in place to make sure nothing breaks, as the two tables behave differently.
- Rows for the virtualized body are fetched in batches of 100 rows from `/ds/get` and stored in local state.
- `ContentLoader` (animated loading state placeholders) are shown when the overall table is first loading and when new rows are being fetched but haven't yet returned from the server.

TODO: the local state used to store fetched rows will grow indefinitely, and could get huge if a user loads an entire dataset with many 100-row api calls.  We will want to limit the size of this local state, and prune rows that are far away from the section the user is viewing.  We can punt on this until it becomes a problem.


src/features/dsComponents/body/BodyTable.tsx
```diff
@@ -1,226 +1,219 @@
...
+import ContentLoader from 'react-content-loader'
 import classNames from 'classnames'
-import ContentLoader from "react-content-loader"
+import useDimensions from 'react-use-dimensions'
+import 'react-virtualized/styles.css'
 
+import { QriRef } from '../../../qri/ref'
 import DataType from '../../../chrome/DataType'
 
+const PAGE_SIZE = 100
```
ramfox: We should rename the `bodyPageSizeDefault` const in `/src/features/dataset/state/datasetActions.ts` to `DEFAULT_BODY_PAGE_SIZE`

Then then we should replace `PAGE_SIZE` with importing `DEFAULT_BODY_PAGE_SIZE`

src/features/dsComponents/body/BodyTable.tsx
```diff
@@ -1,226 +1,219 @@
...
-    this.state = {
-      bodyLength: 0
-    }
+const BodyTable: React.FC<BodyTableProps> = ({
+  rowCount,
+  headers = [],
+  qriRef,
+  loading = false
+}) => {
+  const [ list, setList ] = useState<Record<number, any>>({})
```
ramfox: because we are using local state here, it shouldn't be too difficult to refactor this to use the state tree

Rather than using `list` and `setList` from `useState`, we can pull list using a new`selectBody` selector (that would be defined in `src/features/dataset/state/datasetState.ts` or just add back the `body` props.

rather than using `setList` we should expect that the extended list will get selected from the reducer.

src/features/dsComponents/body/BodyTable.tsx
```diff
@@ -1,226 +1,219 @@
...
+              list[parseInt(startIndex) + i] = row
+            })
+
+            setList({
+              ...list,
+              ...newList
+            })
+
+            resolve()
+          })
```
ramfox: If we refactor `loadBody`'s signature to use limit and offset instead of page and pageSize:

`export function loadBody(ref: QriRef, offset: number = 0, Limit: number = DEFAULT_BODY_PAGE_SIZE): ApiActionThunk`

& refactor `fetchBody` to also use offset and limit:

```
function fetchBody (ref: QriRef, offset: number, limit: number): ApiAction {
  return {
    type: 'body',
    ref,
    [CALL_API]: {
      endpoint: 'ds/get',
      method: 'GET',
      segments: {
        username: ref.username,
        name: ref.name,
        path: ref.path,
        selector: ['body'],
        limit,
        offset
      }
    }
  }
}
```

We _should_ be able to replace the fetch above with 

`loadBody(qriRef, parseInt(stopIndex) - DEFAULT_BODY_PAGE_SIZE + 1)`

ramfox (changes requested):
First. Using this in the app looks fabulous & is a really clean experience!!!

I'm requesting we do some light refactoring to pull the `list` from the component state into the state tree itself!

chriswhong:
Added the requested changes.
- virtualized bodies now use the state tree, actions, and reducers
- `DEFAULT_BODY_PAGE_SIZE` is a constant
- Fixed the dataset header to not say "previewing..." when viewing a full table
- Fixed some UI flashing caused by choosing a cell loading state animation at random

ramfox (approved):
LGTM!

